PR #118, opened 2015-01-15
eager load should find indexes inside engines too

leemhenson:
I'm sure this used to work, but maybe I'm misremembering. If you have a rails app that includes engines and those engines define chewy indexes, they are not found by the rake task when you do ```rake chewy:update:all```. 

What is the reasoning for using this:

```ruby
Rails.application.config.paths['app/chewy'].existent
```

and then globbing further?

### DIFF
--- a/lib/chewy/railtie.rb
+++ b/lib/chewy/railtie.rb
@@ -30,7 +30,8 @@ module Chewy
     end
 
     console do |app|
-      Chewy.logger = ActiveRecord::Base.logger
+      Chewy.logger = ActiveRecord::Base.logger if defined?(ActiveRecord)
+
       if app.sandbox?
         Chewy.strategy(:bypass)
       else
--- a/lib/chewy/railtie.rb
+++ b/lib/chewy/railtie.rb
@@ -1,5 +1,9 @@
 module Chewy
   class Railtie < Rails::Railtie
+    def self.all_engines
+      Rails::Engine.subclasses.map(&:instance) + [Rails.application]
+    end
+
     class RequestStrategy
       def initialize(app)
         @app = app
@@ -54,7 +58,9 @@ module Chewy
     end
 
     initializer 'chewy.add_app_chewy_path' do |app|
-      app.config.paths.add 'app/chewy'
+      Chewy::Railtie.all_engines.each do |engine|
+        engine.paths.add 'app/chewy'
+      end
     end
   end
 end
--- a/lib/tasks/chewy.rake
+++ b/lib/tasks/chewy.rake
@@ -13,8 +13,12 @@ def subscribe_task_stats!
 end
 
 def eager_load_chewy!
-  Rails.application.config.paths['app/chewy'].existent.each do |dir|
-    Dir.glob(File.join(dir, '**/*.rb')).each { |file| require_dependency file }
+  dirs = Chewy::Railtie.all_engines.map { |engine| engine.paths['app/chewy'].existent }.flatten.uniq
+
+  dirs.each do |dir|
+    Dir.glob(File.join(dir, '**/*.rb')).each do |file|
+      require_dependency file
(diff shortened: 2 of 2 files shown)


pyromaniac:
`Rails.application.config.paths['app/chewy'].existent` this supposed to return all existent `app/chewy` paths from every engine loaded. And I'm globbing inside every directory just to make sure namespaced indexes are loaded as well.

Isn't this work? Or maybe you are trying to force this to work outside rails app?

leemhenson:
This is a rails 4.1 app. It's not finding any index files at all.

```
[1] pry(main)> Rails.application.config.paths['app/chewy'].existent
[]
[2] pry(main)> Dir[Rails.root + '**' + 'app' + 'chewy' + '**' + '*.rb']
[
    [ 0] "/<engine path>/app/chewy/foo_index.rb",
    [ 1] "/<engine path>/app/chewy/bar_index.rb"
]
```

:person_frowning:

pyromaniac:
Could you please investigate this further? Your way would not work with engines from gems. I'm pretty sure we have to use `Rails.application.config.paths['app/chewy']` API, but it might be I'm using it just wrong.

pyromaniac:
Wow, this should be used like: `Rails.application.railties.engines.map do |r| r.config.paths.add('app/chewy'); r.config.paths['app/chewy'].existent end`

leemhenson:
that doesn't work >= 4.1 due to:

https://github.com/rails/rails/commit/4a2a504f40a8b6eec88e6aafefcf8e579a24e1c9#diff-7d0a784885a7ba0f99b77cb306dba31b

:grimacing:

pyromaniac:
Oh gosh

pyromaniac:
But there should be the way. Rails got all the engines e.g. app/models directories somehow. We have just to use the same approach.

leemhenson:
```ruby
  engines = if ((Rails::VERSION::MAJOR == 4) && (Rails::VERSION::MINOR >= 1)) || (Rails::VERSION::MAJOR > 4)
    Rails::Engine.subclasses.map(&:instance)
  else
    Rails.application.railties.engines
  end
```

ish

pyromaniac:
Actually we could use `Rails::Engine.subclasses.map(&:instance)` only. We don't have to check anything here

leemhenson:
I think ```instance``` was added here, as part of Rails 4.1:

https://github.com/rails/rails/commit/81b7416afa10934022ad8b1b486419d5a4ed1349

pyromaniac:
Ah, got the difference, then let's just check Rails.application.railties.engines method respondance instead of rails version

pyromaniac:
Look what I've just found: https://github.com/rails/rails/blob/3-2-stable/railties/lib/rails/engine/railties.rb#L27

leemhenson:
This would be a good time for an Ouroboros emoji.

pyromaniac:
And Rails.application is an engine btw, but it is not included in engine list

pyromaniac:
Also could I ask you to perform engines paths adding here: https://github.com/toptal/chewy/blob/master/lib/chewy/railtie.rb#L56 as well

lib/chewy/railtie.rb
```diff
@@ -1,5 +1,15 @@
 module Chewy
   class Railtie < Rails::Railtie
+    def self.all_engines
+      engines = if Rails.application.railties.respond_to?(:engines)
+        Rails.application.railties.engines
+      else
+        Rails::Engine.subclasses.map(&:instance)
+      end
```
pyromaniac: Currently this looks great, but I still don't understand, why do we need this check? It seems `Rails::Engine.subclasses.map(&:instance)` supposed to work with any rails >= 3.2

pyromaniac:
Awesome fix, thanks, man!

Issue #51, opened 2014-08-05
Can i write a Sidekiq job updating the index async?


danielkummer:
Hi there

I'm just getting started with your chewy gem because it looks great and I'd like to try it.
A question. Is it possible to hook into the update_index action and perform the updating inside an async sidekiq job?

Example: 
````
class User <  ActiveRecord::Base
  after_save    { Indexer.perform_async(:index, self.class.name,  self.id) }
  ......
end    
````
And a Sidekiq worker named "Indexer":
````
class Indexer
  include Sidekiq::Worker

  def perform(operation, model_name, record_id)
    #update elasticsearch index here...
  end
end
````

baothi:
when i do again bin/spring stop it run `TacosIndex but when i run TacosIndex::Taco.import(Taco.all)`
it error `Faraday::ConnectionFailed: Connection refused - connect(2) for "localhost" port 9200 `

pyromaniac:
You have to install and run elasticsearch: `brew install elasticsearch` on OS X

baothi:
yes thanks pyromaniac ^^

averell23:
@pyromaniac If you have a generic way of implementing indexing strategies, I'll be happy to fold my code in. The reason I wrote the gem was that the async update was very different to the way chewy handles updates.

Also, what specifically isn't up to date? Even though it's scratchy-ware, we do use the gem in production. So it will get eventually updated when we update chewy and/or Rails. If that's not fast enough for you guys, patches are welcome ;-)

pyromaniac:
Hey @averell23! By not-up-to-date I meant current master for sure. And also https://github.com/toptal/chewy/pull/119 here is the PR you could check out and even try to cooperate :)

dapi:
BTW Use `after_commit` instead of `after_save`. Becase worker can catch job BEFORE database close transaction or transaction can be rollbacked.

pyromaniac:
I've added this change a couple of days ago

PR #119, opened 2015-01-15
Asynchronous strategy for perfoming import in background

sharkzp:
Just proof of concept to decide on what direction @pyromaniac decide to moving.

### DIFF
--- a/chewy.gemspec
+++ b/chewy.gemspec
@@ -30,4 +30,5 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'activesupport', '>= 3.2'
   spec.add_dependency 'elasticsearch', '>= 1.0.0'
+  spec.add_dependency 'sidekiq'
 end
--- a/lib/chewy/config.rb
+++ b/lib/chewy/config.rb
@@ -29,7 +29,10 @@ module Chewy
       # If you neet to return to the previous chewy behavior -
       # just set it to `:bypass`
       #
-      :request_strategy
+      :request_strategy,
+
+      # The SideKiq queue in which all jobs will be saved.
+      :queue
 
     def self.delegated
       public_instance_methods - self.superclass.public_instance_methods - Singleton.public_instance_methods
@@ -41,6 +44,7 @@ module Chewy
       @filter_mode = :and
       @root_strategy = :base
       @request_strategy = :atomic
+      @queue = :chewy_indexer
     end
 
     def transport_logger= logger
--- a/lib/chewy/strategy/asynchronous.rb
+++ b/lib/chewy/strategy/asynchronous.rb
@@ -1,0 +1,20 @@
+module Chewy
+  class Strategy
+    # This strategy has been build on Atomic strategy.
+    # But instead of populating/updating indexes in real time
+    # Here we put workers in a queue and process them in background
+    #
+    # You could customize queue name in your Chewy configuration file
+    class Asynchronous < Atomic
+      def leave
+        @stash.all? do |type, ids|
+          Sidekiq::Client.push(
+            queue: Chewy.sidekiq_queue,
+            class: Chewy::Worker,
+            args:  [type, ids]
+          )
+        end
+      end
+    end
+  end
+end
--- a/lib/chewy/worker.rb
+++ b/lib/chewy/worker.rb
@@ -1,0 +1,10 @@
+module Chewy
+  class Worker
+
+    include Sidekiq::Worker
+
+    def perform(index_name, ids)
+      index_name.import ids
+    end
+  end
+end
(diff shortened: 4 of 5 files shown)


pyromaniac:
Several points about this:
1. I definitely don't want chewy to be dependent on any delayed jobs gem. So Async strategy supposed to be optional and have to load all the classes on-demand when it's used in application.
2. I think there should be 2 kind of async strategies: urgent and atomic, working the same way as existent, but async.
3. Sidekiq strategies supposed to be named like `sidekiq_urgent` and `sidekiq_atomic` respectively.

pyromaniac:
Also maybe it is possible to make one option like `async_config` to accept just a hash which could be passed to the particular worker. Or maybe own option for every async executor such as `sidekiq_options` or `delayed_job_options` in future. Have to think about it.

sharkzp:
@pyromaniac okay sounds resonable. 
What if we do this:
1. On a startup/initialization of a gem we will find out what app support(Sidekiq, DelayJob, Resque).
2. Depending on this we will wrap `import` call to a proper asynch gem.

Than we could do two things.
1. Create `asynch_update_index`
2. Create asynch strategy that will wrap everything to a worker.

pyromaniac:
There might be cases when Sidekiq and DJ are working together in one project, even temporarily, but we have to give an ability to choose the backend, that's why I'm suggesting to use separate strategies.

And I personally don't like the idea of `asynch_update_index`, there should be only one way to define updates. And strategy defines which way update happens

sharkzp:
I'm a bit busy currently but will resume work on this in a few days.

averell23:
Hi. As I mentioned in #51 - I already wrote a gem to run chewy jobs async. As we use that in Production, we already ran into some issues that we fixed and that are generally valid. E.g.:

* The async jobs **must** be queued on "after_commit" (not after update or save!) - otherwise it is not ensured that the data has been committed to the database when the job starts. This is generally true for **all** sidekiq jobs that use the database, but I am not sure that it's a useful default behaviour for chewy.
* We found it neccessary to implement locking in the job, so that no two jobs can update the same model object at the same time. Otherwise there will be race conditions that lead to unexpected side effects.

I'm open to cooperate and change my code to be a "standard" strategy. There are several remarks and reasons we haven't done so yet:

* Dependency on Sidekiq and related gems: Even if we have an "official" strategy, I would very much prefer to bundle it into a separate plug-in gem, instead of automagically detecting stuff by default. We'd need to agree on a mechanism how to use plugins with chewy.
* The "atomic" strategy needs to be used inside a block, and doesn't automatically enforce the update at some point in time. The "chewy solution" is to wrap this into the application controller - but many model updates are not run inside a web action (but from a rake task etc.). Also, since we use after_commit, we will get only one call per database transaction anyway, and for background jobs the problem of multiple updates is a lot less urgent.
* Using after_commit breaks with Rails' default transactional fixtures (and any database cleaning in tests that uses transactions). You need to inject a bit of hackiness to fix this.

Best, Daniel

pyromaniac:
So, several points on it:

* Chewy 0.7.0 was released, so async strategy would definitely be included at the next release.
* Chewy now utilizes after_commit for AR
* The problem with transactional fixtures was solved by optional using of after_save callbacks (I personally think it should be fallback on the AR lever btw)
* About rake tasks. I should say that strategy should be chosen for every particular rake task (or every models updating place outside the controller) separately and reasonably. There are a lot of cases when bypass is much more preferable in my practice.
* after_commit doesn't mean only one callback call. It called for every objects the same way as after_save, but after commit only, so we still need atomic strategy to avoid multiple and duplicated index requests
* I came to the conclusion that async urgent strategy is excessive, so we need only atomic analogue.
* This PR is almost ok, but I would move sidekick job inside the strategy module, rename strategy to `sidekick` and remove the gem dependency.

pyromaniac:
Closing this since the feature is implemented in maser, thanks guys! And could you please find a minute to check it out?

averell23:
Just for clarification: after_commit means one callback call _per transaction_ (each transaction commits exactly once). Since we usually wrap important updates in transactions, that worked reasonably well for us. Of course calling #save multiple times outside a transaction will still do multiple callbacks.

averell23:
Also, I'll check out the new code, of course, and deprecate my gem if we can use it. What about the idea of making those strategies plugable as external gems, though?

pyromaniac:
Yeah, I meant the case with several objects. The flow (for 2 objects) will be: 

```
after_save1
after_save2
commit
after_commit1
after_commit2
```

pyromaniac:
Yeah, I can implement a kind of strategies repository, but currently anybody is able to add new strategy as a gem, just following class name convensions

PR #121, opened 2015-01-29
Allow scrip_score to receive params

joeljunstrom:
By passing params scripts can now be cached even when we want to pass values to them.

Unfortunately I have a very limited grasp on both ElasticSearch and Chewy so please advise if this is the wrong way to go about it or if this is already implemented somewhere (I couldn't find it).

### DIFF
--- a/lib/chewy/query.rb
+++ b/lib/chewy/query.rb
@@ -348,21 +348,25 @@ module Chewy
     # added to the search request and combinded according to
     # <tt>boost_mode</tt> and <tt>score_mode</tt>
     #
-    #   UsersIndex.script_score("doc['boost'].value", filter: { term: {foo: :bar} })
+    #   UsersIndex.script_score("doc['boost'].value", params: { modifier: 2 }, filter: { term: {foo: :bar} })
     #       # => {body:
     #              query: {
     #                function_score: {
     #                  query: { ...},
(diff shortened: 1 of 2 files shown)


pyromaniac:
I think this is just a wrong merging way implemented here, it should be more like: `scoring = {script_score: { script: script }.merge(options))`

joeljunstrom:
Yea that was my first guess but since it was very explicit about setting `script_score: { script: script }` and thus having the rest of the keys on the level above I figured it was by design (elasticsearch seems to have some funky ways to describe queries)

joeljunstrom:
The comment (https://github.com/toptal/chewy/blob/master/lib/chewy/query.rb#L360) also explicitly places the filters key in the top level which lead me to that conclusion.

joeljunstrom:
Updated the implementation, if this is the expected functionality I'll rebase it.

joeljunstrom:
Sorry for the comment spam here… Looking at http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html it does seem like `params` is the only key other than `script` that resides within the `script_score` object. There is however an example where a function contains both a `script_score` and a `filters` object.

pyromaniac:
Awesome fix, thanks!

PR #123, opened 2015-02-03
Add failing test case for urgent strategy and default scope

gkop:
This is an "interesting" regression I noticed since the work refactoring the strategies.

We have different behavior between urgent and atomic strategy: atomic strategy passes id backrefs, while urgent strategy passes the objects themselves.

This ends up causing the atomic strategy updates to respect the default activerecord scope, while the urgent strategy updates do not: https://github.com/toptal/chewy/blob/0c726fb79912817e662ac432ead978ad33fce88a/lib/chewy/type/adapter/active_record.rb#L74-L80 .

The easiest fix might be only using id backrefs in this case in ActiveRecord#import, but the existing test suite supports indexing records for which id is nil, so I cannot make this fix on my own since it will invalidate many existing tests.

Also it occurred to me to repeat the default scope logic when passing the backref in the model observing block, but this doesn't work: in this block we can only decide whether or not to update the index for the backref, not whether to reindex or delete. So if a record goes from in scope to out of scope, we must be observing it in order to fulfill the necessary deletion.

So finally I am left with the fix I propose in this PR, modifying the urgent strategy. It's not a very satisfying fix since it doesn't address the root of the problem. But perhaps it's better than the current inconsistency between strategies.

### DIFF
--- a/lib/chewy/strategy/urgent.rb
+++ b/lib/chewy/strategy/urgent.rb
@@ -10,7 +10,12 @@ module Chewy
     #
     class Urgent < Base
       def update type, objects, options = {}
-        type.import(objects)
+        objects = Array(objects)
+        if objects.all? { |object| object.respond_to?(:id) }
+          type.import(objects.map(&:id))
+        else
+          type.import(objects)
+        end
       end
     end
   end
--- a/lib/chewy/strategy/urgent.rb
+++ b/lib/chewy/strategy/urgent.rb
@@ -10,7 +10,7 @@ module Chewy
     #
     class Urgent < Base
       def update type, objects, options = {}
-        objects = Array(objects)
+        objects = Array.wrap(objects)
         if objects.all? { |object| object.respond_to?(:id) }
           type.import(objects.map(&:id))
         else
--- a/spec/chewy/strategy_spec.rb
+++ b/spec/chewy/strategy_spec.rb
@@ -74,5 +74,23 @@ describe Chewy::Strategy do
         Chewy.strategy(:atomic) { [city, other_city].map(&:save!) }
       end
     end
+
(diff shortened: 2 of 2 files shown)


gkop:
Re: mongoid, I ran `respec -t mongoid` to try to replicate the issue in travis but got a lot of strange errors:

```
     Failure/Error: specify { expect(PlacesIndex.limit(6).load(city: { scope: ->{ where(:rating.lt => 2) } }).order(:rating))
     NoMethodError:
       undefined method `to_sym' for #<Origin::Key:0x0000000680ff88>
```

If you want to eventually merge this test case in, I can surely make it work with mongoid's named scopes.

pyromaniac:
Hey, thanks for the patch, but I have to think about it. Actually, it was never supposed to use conditions in default scope. Instead there is `delete_from_index` method which might be defined in model to specify such conditions.

After your PR I've got 2 thoughts:
1. It would be more convenient to define delete_from_index or similar method not in model, but in chewy type since there might be multiple types defined for one model.
2. Chewy should extract conditions from default scope to use them exactly the same way as delete_from_index method, but this would work in simple cases only.

So thanks for pointing on the problem, I'll think about it a little bit.

lib/chewy/strategy/urgent.rb
```diff
@@ -10,7 +10,12 @@ class Strategy
     #
     class Urgent < Base
       def update type, objects, options = {}
-        type.import(objects)
+        objects = Array(objects)
```
pyromaniac: Could I ask you to replace this with Array.wrap ?

pyromaniac:
I've decided to force import to respect default type scope, so this PR is not relevant anymore. Could you try master please?

PR #127, opened 2015-02-05
[Fix doc] s/types_hash/type_hash

amerov:
UsersIndex.types_hash
NoMethodError: undefined method `types_hash' for UsersIndex


pyromaniac:
Thanks

PR #144, opened 2015-03-23
Allow to use scripting in the search criteria

ka8725:
The feature can be useful to get a value on the search result which is calculated via scripts. It may be calculation distance between geopoints, for example.


pyromaniac:
Awesome, thanks!

Issue #148, opened 2015-03-24
What about search_type?


marshall-lee:
Sometimes i need to fetch only aggregations(sum aggs, term aggs, etc).
It could be done with `search_type: 'count'` but there's no option in Chewy to do it.

Now i use this workaround:

```ruby
query.send(:_request).merge!(search_type: 'count')
```

It really works but need a native support for this.

pyromaniac:
Good point, will add this some day, but actually you can send a small PR and this would be much better and faster.

marshall-lee:
I thought about it but I didn't find a good place in code for this. Okay I will look for this soon.

PR #149, opened 2015-03-25
Add search_type support

marshall-lee:
This one is for #148

### DIFF
--- a/lib/chewy/query.rb
+++ b/lib/chewy/query.rb
@@ -818,6 +818,20 @@ module Chewy
       chain { criteria.update_types params, purge: true }
     end
 
+    # Sets <tt>search_type</tt> for request.
+    # For instance, one can use <tt>search_type=count</tt> to fetch only total count of records or to fetch only aggregations without fetching records.
+    #
+    #   scope = UsersIndex.search_type(:count)
+    #   scope.count == 0  # no records actually fetched
+    #   scope.total == 10 # but we know a total count of them
+    #
+    #   scope = UsersIndex.aggs(max_age: { max: { field: 'age' } }).search_type(:count)
+    #   max_age = scope.aggs['max_age']['value']
+    #
+    def search_type val
+      chain { options.merge!(search_type: val) }
+    end
+
     # Merges two queries.
     # Merges all the values in criteria with the same rules as values added manually.
     #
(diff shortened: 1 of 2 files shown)


marshall-lee:
https://travis-ci.org/toptal/chewy/jobs/55774252 fails but because of improper ordering of database records.

lib/chewy/query.rb
```diff
@@ -818,6 +818,20 @@ def types! *params
...
+    #
+    #   scope = UsersIndex.search_type(:count)
+    #   scope.count == 0  # no records actually fetched
+    #   scope.total == 10 # but we know a total count of them
+    #
+    #   scope = UsersIndex.aggs(max_age: { max: { field: 'age' } }).search_type(:count)
+    #   max_age = scope.aggs['max_age']['value']
+    #
+    def search_type val
+      chain { options.merge!(search_type: val) }
```
pyromaniac: Could you please use `update_options` method here?

marshall-lee: @pyromaniac okay but before it please look at this one: #150, i don't want this build to fail :smiley:

marshall-lee: Are you sure about `criteria.update_options`? I think that `Criteria` is just responsible for constructing a `{ body: ... }` part but `search_type` is not a part of body. It's just a query option just like `type` or `index`.
Or you're suggesting to add a new `update_options` method to `Query` ?

pyromaniac: https://github.com/toptal/chewy/blob/master/lib/chewy/query/criteria.rb#L43
https://github.com/toptal/chewy/blob/master/lib/chewy/query.rb#L143

This is just more cmmon way to do the same thing

marshall-lee: It's not the same thing! `criteria.update_options` updates criteria's options, not query's options.

marshall-lee: It's not possible to just `chain { criteria.update_options search_type: val }`. It requires me to move `search_type` logic to `Criteria` class. I don't think that search_type is a Criteria's responsibility. It is `Query`'s responsibility.

pyromaniac: Oh, got it now. Not sure if I'm completely agree with you, but currently it is ok.

Issue #164, opened 2015-05-02
Expose the mapped model class to the block passed to define_type


caldwecr:
Consider ...

```
module CustomTypeExtension
  def extended_field(*args, &block)
    # I want to be able to know that the Index document type is mapped to the class Models::Foo
    self.underlying_model.do_something #this method doesn't really exist
    field *args, &block
  end
end

class FooIndex < Chewy::Index
  define_type Models::Foo do
    extend CustomTypeExtension
    extended_field :cat, :dog
  end
end
```

I'd like to be able to access the associated Model's class inside of the block passed to define_type in general, and inside of mixed in modules specifically. I went trolling through the methods but I couldn't find anything appropriate. Advice?

caldwecr:
I'm happy to submit a PR that accomplishes this if you let me know where you'd like to see the method which exposes the model class added and what you think it should be named.

pyromaniac:
`adapter.target` actually

caldwecr:
Thank you.

PR #165, opened 2015-05-05
Update README with a section about the scope inside of define_type

caldwecr:
Related to issue #164 it seemed prudent to create a section of the README that outlines important aspects of the scope inside of a block that is passed to define_type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,14 @@ See [config.rb](lib/chewy/config.rb) for more details.
   end
   ```
 
+### Inside the scope of .define_type
+
+You can access the class that the defined type is mapped to using:
+
+```ruby
+adapter.target
+```
+
 ### Types access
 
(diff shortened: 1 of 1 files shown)


caldwecr:
Confused why this build failed. It's a change to the readme only.

pyromaniac:
Just a random fail, never mind. Also, I'm not sure we need this info in the readme, this is quite advanced and obvious fact for those who decided to dig sources. How did you come to this? Why did you need to access target model?

caldwecr:
I agree that is is a pretty advanced piece of knowledge. The scenario I encountered was that I was authoring a module to be mixed into the created class. It was (at the time) non-obvious how I could reference the target model from the mixed in module. I was preparing to add the note to our internal documentation and thought it might be beneficial to others.

I have no qualms with disregarding this proposed change.

PR #172, opened 2015-05-13
'load_options' assignement is never used

waleo:


pyromaniac:
Yeah, it is never used, but I would like to leave it as is just in case.

waleo:
Okay.

pyromaniac:
Thanks for your attention to project

Issue #179, opened 2015-05-20
Rspec config for Chewy.use_after_commit_callbacks


ka8725:
Using RSpec 3.
Figured out that independently on the `Chewy.use_after_commit_callbacks` variable change in `before` callbacks and even before `RSpec.configure` the value is not updated in callbacks context to update index. So I got the following head of `spec_helper.rb` file:

```ruby
ENV['RAILS_ENV'] ||= 'test'

require 'chewy'
require 'chewy/rspec'
Chewy.use_after_commit_callbacks = false

require File.expand_path('../../config/environment', __FILE__)
require 'rspec/rails'
```

The `Chewy.use_after_commit_callbacks` must be set before all models' classes are initialized. May be add this to README or post to some wiki page?

pyromaniac:
Feel free to do this. Moreover, if you are using spring models are loaded even before spec_helper.rb, so you have to create an initializer with `Chewy.use_after_commit_callbacks = false if Rails.env.test?`

ka8725:
`Chewy.use_after_commit_callbacks = false if Rails.env.test?` doesn't work for me. Using spring, models' classes are already initialized in an initializer.

pyromaniac:
Something wrong with your application, you are calling models in initializers or something like this. Normally models are loaded after initializers

ka8725:
@pyromaniac , done - https://github.com/toptal/chewy/pull/184

PR #181, opened 2015-05-20
Add note to README about rspec and transactions

cymen:
We recently ran into an issue with testing when we started using Database Cleaner and it took a little while to track down: testing for `update_index` to have happened when using an rspec wrapper that makes uses of transactions will fail. Instead, one needs to use truncation or some other method around the suite that is testing the index updates.


pyromaniac:
Actually there is another option: https://github.com/toptal/chewy/issues/179

cymen:
@pyromaniac Great, it would be nice to have that documented too.

ka8725:
Can be closed because of this: https://github.com/toptal/chewy/pull/184

pyromaniac:
Done, thanks

PR #185, opened 2015-05-28
.import(nil) should be treated as equivalent to .import()

caldwecr:
Encountered this issue when trying to implement .deep_import

I attempted to implement deep_import without using the splat'd args approach and nil as a default for the first arg. This seemed to work fine except that when the default was passed through Chewy treated it as `nil == []` (In truth it may actually have treated it as the nil scope and not the empty array).

Not sure what you (@pyromaniac) view as the correct behavior. If you agree that .import(nil) should be treated as .import() then this PR will resolve same. If you disagree and feel that .import(nil) should continue to be treated as .import([]), please let me know and I'll open a separate PR that adds comments and specs to reflect the intent.

Currently .import() imports the default data set and .import([]) imports the empty data set.
Currently .import(nil) imports the empty data set.

This PR changes the behavior to be equivalent to the .import() method

### DIFF
--- a/lib/chewy/type/adapter/orm.rb
+++ b/lib/chewy/type/adapter/orm.rb
@@ -77,7 +77,7 @@ module Chewy
           import_options = args.extract_options!
           batch_size = import_options[:batch_size] || BATCH_SIZE
 
-          collection = args.empty? ? default_scope :
+          collection = import_all?(args) ? default_scope :
             (args.one? && args.first.is_a?(relation_class) ? args.first : args.flatten.compact)
 
           if collection.is_a?(relation_class)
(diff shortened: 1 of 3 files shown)


caldwecr:
Looking at the failure output - which seems to come from the AR adapter perhaps your intent was that import(nil) is equivalent to import([])

pyromaniac:
Yeah, that was fixed not so long ago (https://github.com/toptal/chewy/commit/3509b04644d2ab3cd13d73fedd030edba3b5b62b) since it was causing unexpected problems. I think current behavior is correct. So use splat please if you need.

PR #188, opened 2015-06-01
Accept table name for update_index as a Proc

SeTeM:
Hello! I'm using "One Language per Document" pattern (https://www.elastic.co/guide/en/elasticsearch/guide/current/one-lang-docs.html) and I need to define types dynamically. My PR will help with this issue.

Example:
```ruby
class BooksIndex < Chewy::Index
  %w(ru en).each do |language|
    define_type Book.where(language: language), name: "book_#{language}" do
      field :title, type: 'string', analyzer: language
    end
  end
end

class Book < ActiveRecord::Base
  update_index(->(book) {"books#book_#{book.language}"}) { self }
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,12 @@ See [config.rb](lib/chewy/config.rb) for more details.
     update_index('users') { users } # if index has only one type
                                     # there is no need to specify updated type
   end
+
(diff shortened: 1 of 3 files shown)


pyromaniac:
Awesome job, thanks!

PR #193, opened 2015-06-10
Correct obvious mistake in comments of query

caldwecr:
Replace 'Deletes' with 'Find' in the .find method's comment block.


pyromaniac:
Thanks!

PR #204, opened 2015-06-29
Encapsulate rake helper methods

caldwecr:
This PR moves existing rake helper method logic into an isolated module / file.

### DIFF
--- a/lib/tasks/chewy.rake
+++ b/lib/tasks/chewy.rake
@@ -1,31 +1,30 @@
 require 'tasks/rake_helper'
 namespace :chewy do
-  include RakeHelper
   desc 'Destroy, recreate and import data to specified index'
   task :reset, [:index] => :environment do |task, args|
-    subscribe_task_stats!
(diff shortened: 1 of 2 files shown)


caldwecr:
Please don't merge this yet. I believe these methods may not be covered by specs. I'd like to see that resolved before the merge as there's no suitable way to test without it.

pyromaniac:
Ok, how about extend self and use method directly in rake without inclusion?

caldwecr:
@pyromaniac - sure; just to be clear you want me to turn RakeHelper into a singleton and expose the methods on the module directly, right? And then invoke them in the rake task as RakeHelper.method_name

pyromaniac:
A kind of, but I'm just offering since I've got no opinion on this, so don't hesitate to choose the best approach and implement it :)

caldwecr:
Ok @pyromaniac I think this should be good.

pyromaniac:
Thanks!

PR #207, opened 2015-06-30
Add adapter for Sequel

jirutka:
I’ve added adapter for [Sequel](https://github.com/jeremyevans/sequel) and also refactored adapter selection. Sorry for missing specs, I don’t have a time for it now. :(


pyromaniac:
Great contribution!
Seems like you forgot to map Chewy::Type::Adapter::Object
Also are you going to add specs in future or offering somebody else to do it?

jirutka:
> Seems like you forgot to map Chewy::Type::Adapter::Object

Yeah, I’ve already fixed it.

> Also are you going to add specs in future or offering somebody else to do it?

I’ve eventually prepared configuration for testing sequel adapter and ported some specs from active_record adapter. The rest of specs (`#import`, `#load`) are, to be honest, quite crazy, so I would like to let it for someone else.

pyromaniac:
Fair enough, I'll do it for you as soon as current code will pass existing specs

jirutka:
All tests are finally green! :)

pyromaniac:
Such an awesome piece of work.

But I'm looking on it and thinking: are you sure that import and load works as they supposed to? I mean, I can merge it right now and simply forget about it for a time because I'm not sure I'll have a time to write such a specs in the near future, but there will be probably broken, or at least, unverified by specs adapter and even you are going to use it without being 100% certain that it works correctly. I see that while things are moving you've fixed a lot of stuff here and who knows how many pitfalls could there be in this code yet.

jirutka:
I’m quite sure that import and load works, because the most code is already tested with existing specs. Most of the specs are more integration than unit tests, so I found many bugs in sequel adapter and observer plugin thanks to them. I’m not so sure with [these four lines](https://github.com/jirutka/chewy/blob/ccb9958bf17d63fe126be1d9921b0893cc564e75/lib/chewy/type/adapter/sequel.rb#L66-L69) though, because they are not covered by any spec. However, this feature isn’t critical.

jirutka:
Ah, wait a moment, there’s one thing I’ve forgotten to fix (we shouldn’t assume constant `DB`).

jirutka:
Okay, now it’s ready.

pyromaniac:
Alright, let's try it :) Thanks for your contribution.

PR #208, opened 2015-06-30
Add tasks.rb to simplify loading rake tasks

jirutka:
Now you can load chewy rake tasks in the application's Rakefile like this:

    require 'chewy/tasks'


pyromaniac:
Since rails is loading this tasks with railtie, which framework are you using?

jirutka:
I’m not using Rails, but [Pliny](https://github.com/interagent/pliny). However, I’ve eventually found that these rake tasks depends on Rails, so I can’t use them anyway.

pyromaniac:
Probably it make sense to use Rails part separately

PR #209, opened 2015-06-30
Travis: enable containerized (faster) builds

jirutka:


pyromaniac:
Thanks! Builds seems to be faster indeed

PR #212, opened 2015-07-05
Require only core extensions which are actually used

jirutka:
The aim is to reduce monkey patching of core classes which is undesirable, especially when this gem is used outside of Rails app.


pyromaniac:
Great, thanks!

Issue #214, opened 2015-07-09
Importing index generates queries with duplicate IDs


giovannibonetti:
Hi! First of all, thanks for this library, it has been very useful in my company!
I think it is better if I explain the issue with my use case:

```ruby
class ServicesIndex < Chewy::Index
  define_type define_type Service.includes(service_template: [], service_category: {category_template: [], business: {pictures: [], week_schedule: []} }, skills: {employment: {user: [], week_schedule: :shifts}}) do
    field :id, :name
    ...
  end
```
```ruby
pry(main)> ServicesIndex.import
SELECT  DISTINCT "services"."id", "services"."id" AS alias_0 FROM "services" LEFT OUTER JOIN "service_templates" ON "service_templates"."id" = "services"."service_template_id" LEFT OUTER JOIN "service_categories" ON "service_categories"."id" = "services"."service_category_id" LEFT OUTER JOIN "category_templates" ON "category_templates"."id" = "service_categories"."category_template_id" LEFT OUTER JOIN "businesses" ON "businesses"."id" = "service_categories"."business_id" LEFT OUTER JOIN "pictures" ON "pictures"."depicted_id" = "businesses"."id" AND "pictures"."depicted_type" = $1 LEFT OUTER JOIN "week_schedules" ON "week_schedules"."schedulable_id" = "businesses"."id" AND "week_schedules"."schedulable_type" = $2 LEFT OUTER JOIN "skills" ON "skills"."service_id" = "services"."id" LEFT OUTER JOIN "employments" ON "employments"."id" = "skills"."employment_id" LEFT OUTER JOIN "users" ON "users"."id" = "employments"."user_id" LEFT OUTER JOIN "week_schedules" "week_schedules_employments" ON "week_schedules_employments"."schedulable_id" = "employments"."id" AND "week_schedules_employments"."schedulable_type" = $3 LEFT OUTER JOIN "shifts" ON "shifts"."week_schedule_id" = "week_schedules_employments"."id"  ORDER BY "services"."id" ASC LIMIT 1000  [["depicted_type", "Business"], ["schedulable_type", "Business"], ["schedulable_type", "Employment"]]
[...]

PR #215, opened 2015-07-09
Fix duplicate ids on index import

giovannibonetti:
This solves https://github.com/toptal/chewy/issues/214 (at toptal/chewy)


pyromaniac:
Seems like it supposed to be working. Thanks!

PR #219, opened 2015-07-15
fix-duplicate-ids-on-index-import -> now with better performance

giovannibonetti:
Hi again!
Here in my company we are relying heavily on Chewy right now, and we are looking for performance bottlenecks, since we are using some complex views in the database right now.

Regarding https://github.com/toptal/chewy/issues/214, we saw that still we could make it faster if we somehow could remove all the OUTER JOINS that come up in the query as a result of eager loading with Active Record's include. So I asked around at StackOverflow (http://stackoverflow.com/questions/31436175/in-an-activerecord-relation-with-eager-loading-how-can-you-pluck-the-primary-co#comment-50842906), and someone came up with a way to do just that.

Do you think there might be any unintended side-effect because of that? I'm still waiting for the tests to pass...

### DIFF
--- a/lib/chewy/type/adapter/active_record.rb
+++ b/lib/chewy/type/adapter/active_record.rb
@@ -38,7 +38,8 @@ module Chewy
         end
 
         def pluck_ids(scope)
-          scope.pluck("distinct(#{target.table_name}.#{target.primary_key})")
+          scope.dup.tap{|r| r.includes_values = []}
+            .pluck("distinct(#{target.table_name}.#{target.primary_key})")
         end
 
         def scope_where_ids_in(scope, ids)
--- a/lib/chewy/type/adapter/active_record.rb
+++ b/lib/chewy/type/adapter/active_record.rb
@@ -38,8 +38,7 @@ module Chewy
         end
 
         def pluck_ids(scope)
-          scope.dup.tap{|r| r.includes_values = []}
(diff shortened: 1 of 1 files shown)


pyromaniac:
It will probably be better to use `.only(:where)` here instead of hacks

pyromaniac:
And maybe use `.uniq.pluck()` since it should generate distinct statement (however, I'm not sure about rails 3)

giovannibonetti:
The problem with `.only(:where)` is that it removes inner joins which might me used for restricting the query:
```ruby
Post.joins(:author).where(author: {name: 'John'}).includes(:comments).only(:where).pluck(:id)
# PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "authors"
```

pyromaniac:
Ok, `.except(:includes)` then.

giovannibonetti:
Updated the `.except(:includes)` part.
Waiting to hear about the `.uniq.pluck()` part.

pyromaniac:
http://guides.rubyonrails.org/v3.2.9/active_record_querying.html#pluck

giovannibonetti:
Now I have just updated the `.uniq.pluck()` part, too.
Back to how it was previously
```ruby
scope.except(:includes).uniq.pluck(target.primary_key.to_sym)
```

pyromaniac:
Looks awesome, waiting for specs, also going to merge this tomorrow already :)

PR #227, opened 2015-08-03
Add condition options to `update_index` macro

barthez:
Options `:if` and `:unless` are just passed to underlying callback definitions in mongo/active_record/sequel.


pyromaniac:
YaY!

silva96:
Why there is no documentation about this?

PR #229, opened 2015-08-04
Fix pluck_ids method for ActiveRecord adapter

barthez:
Sometimes relations has default_scope with order clause that depends on
related model (usually with :through). `pluck` selects only one column, so
columns in order clause is missing. Reordering with no order does the trick.


pyromaniac:
Why not `.except(:includes, :order)` ?

pyromaniac:
Thanks!

PR #231, opened 2015-08-07
Revert "Fix pluck_ids method for ActiveRecord adapter"

barthez:
Reverts toptal/chewy#229

It breaks import! With no order batch load fails.


pyromaniac:
I just didn't get why specs didn't caught this. Could you maybe cover this please?

giovannibonetti:
Maybe the problem has come with a model that uses UUID as the primary key? In that case, I think explicit ordering is necessary (for example, by "created_at") if you have to limit results.

PR #238, opened 2015-08-19
In rake tasks, only search Rails engines with defined app/chewy folders

mattmenefee:
When running the "update all" or "reset all" rake tasks, Chewy would
search all "app/chewy" folders for each Rails engine for indexes to
update or reset. However, if a Rails engine did not have an app/chewy
folder defined, the rake task would fail. This change compacts the
folder list to first remove the list of Rails engines that do not have
the folder defined, before checking to see if there are any existing
indexes in those folders.

### DIFF
--- a/lib/chewy/rake_helper.rb
+++ b/lib/chewy/rake_helper.rb
@@ -17,7 +17,7 @@ module Chewy
       end
 
       def eager_load_chewy!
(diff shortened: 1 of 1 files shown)


pyromaniac:
well, let's leave uniq there as well

mattmenefee:
Ah, good catch. Fixed.

pyromaniac:
Thanks!

Issue #192, opened 2015-06-10
Plan for promoting aggregations to first class


caldwecr:
Looking ahead to ES 2.0 and reducers, and considering our own work with automatically converting parts of the chewy filter DSL directly into aggregations, I'm taking this chance to reach out and ask: have you thought yet about the approach you want to take? If any?

I'd love to see more convenient methods for parameterization of aggregations (currently missing in the elastic search gem client), aggregation side channel integration into the query results, and direct support in the chewy filter DSL for reductions of aggregations. 

I'm sure I can come up with ten more important improvements that I'd love to see, but my intent with this message is to get a feel for your own intent. 

Kind regards. 
Sent from tiny keyboard

pyromaniac:
Frankly, I didn't have a chance to work with aggregations closely, that's why this part of chewy is not elaborated enough, and unfortunately currently I've got no ideas about it. But I'm open for a discussion, so let's try to design this interfaces and implement it then. I definitely agree this should be done.

caldwecr:
Thinking about this more over the last couple of days, I think the two highest impact areas for updating Chewy with regards to aggregations are the ability to _bind aggregate results_ (from the side channel) to query results and a way to _write aggregations that are human readable_ (they probably would reside within a define_type block).

Before I get too far into spec'ing an interface for those two pieces, I want to knock out the less beneficial but more tractable 'deep import' problem. (opening a separate issue to present proposed interface.)

caldwecr:
This comment is WIP on the topic of binding aggregate results.
```
define_type Patient do
  root parent: "physician", parent_id: -> { physician_id } do
    field :name
    field :gender
    field :dob
  end
end

define_type Physician do
  
end
```

caldwecr:
```
class StoreIndex < Chewy::Index
  define_type Product do
    root parent: "store", parent_id: -> { store_number } do
      field :product_name
      field :sku, type: 'integer'

      field :categories do
        field :category_id
        field :category_name
      end
    end
  end

  define_type Store do
    field :store_number, type: 'integer'
    field :categories do
      field :category_id
      field :category_name
    end
  end
end
```

caldwecr:
Lots more coming. But some work in progress for thought.

```
class StoreIndex < Chewy::Index
  define_type Product do
    root parent: "store", parent_id: -> { store_number } do
      field :product_name
      field :sku, type: 'integer'
      field :manufacturer

      field :parts_warranty_in_days, type: 'integer'
[...]

PR #242, opened 2015-08-22
Named aggs

caldwecr:
Adds support for named aggregations as well as fully qualified named aggregations.

### DIFF
--- a/lib/chewy/type/mapping.rb
+++ b/lib/chewy/type/mapping.rb
@@ -118,6 +118,35 @@ module Chewy
           end
         end
 
+        # Defines an aggregation that can be bound to a query or filter
+        #
+        #   Suppose that a user has posts and each post has ratings
+        #   avg_post_rating is the mean of all ratings
+        #
+        #   class UsersIndex < Chewy::Index
+        #     define_type User do
+        #       field :posts do
+        #         field :rating
+        #       end
+        #
+        #       agg :avg_rating do
+        #         { avg: { field: 'posts.rating' } }
+        #       end
+        #     end
+        #   end
+        def agg *args, &block
+          options = args.extract_options!
+          build_root unless root_object
+
+          @_agg_defs ||= {}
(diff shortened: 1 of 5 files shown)


caldwecr:
Related to #192

lib/chewy/type/mapping.rb
```diff
@@ -118,6 +118,35 @@ def field *args, &block
...
+        #         { avg: { field: 'posts.rating' } }
+        #       end
+        #     end
+        #   end
+        def agg *args, &block
+          options = args.extract_options!
+          build_root unless root_object
+
+          @_agg_defs ||= {}
+          @_agg_defs[args.first] = block
```
pyromaniac: Why don't use class_attribute here to make it work correctly in case of inheritance?

caldwecr: Great feedback. Change made. @pyromaniac

caldwecr:
looks g2g

lib/chewy/type/mapping.rb
```diff
@@ -118,6 +119,35 @@ def field *args, &block
...
+        #         { avg: { field: 'posts.rating' } }
+        #       end
+        #     end
+        #   end
+        def agg *args, &block
+          options = args.extract_options!
+          build_root unless root_object
+
+          self._agg_defs ||= {}
+          _agg_defs[args.first] = block
```
pyromaniac: Well, it still modifies parent hash, you have to do `self._agg_defs = _agg_defs.merge(args.first => block)`
And also move `self._agg_defs = {}` near `class_attribute :_agg_defs`. And probably you don't need `def agg_defs` anymore

caldwecr:
@pyromaniac ok now?

pyromaniac:
Give me some time please, I have to review this

pyromaniac:
Ok, Let's see how would it go. I'm still not sure that named scopes can't be used instead of it, but I have at least to try to use aggregations at first

PR #247, opened 2015-09-09
Make it faster

biow0lf:
In this code we don't need position on text, only matching.

Benchmark code: https://gist.github.com/biow0lf/e4cd0082899c2d83f3fe

```
igor@Zubkov-PC:~/opensource/chewy$ ruby benchmark.rb 
       user     system      total        real
  43.820000   0.020000  43.840000 ( 47.276215)
  20.930000   0.020000  20.950000 ( 22.135514)
igor@Zubkov-PC:~/opensource/chewy$ 
```

### DIFF
--- a/lib/chewy/railtie.rb
+++ b/lib/chewy/railtie.rb
@@ -10,7 +10,7 @@ module Chewy
       end
 
       def call(env)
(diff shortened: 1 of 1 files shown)


pyromaniac:
Thanks

PR #253, opened 2015-09-17
Add Query#objects

syndbg:
Added an easier to use `Query#objects` method.

It does the same thing as `UsersIndex.filter(range: {rating: {gte: 100}}).preload(...).query(...).map(&:_object)` - https://github.com/toptal/chewy#objects-loading

But now it can be done faster as `UsersIndex.filter(range: {rating: {gte: 100}}).query(...).objects`

### DIFF
--- a/lib/chewy/query.rb
+++ b/lib/chewy/query.rb
@@ -39,6 +39,7 @@ module Chewy
       @_indexes = indexes_or_types_and_options.select { |klass| klass < Chewy::Index }
       @_indexes |= @_types.map(&:index)
       @criteria = Criteria.new
+      @objects = []
     end
 
     # Comparation with other query or collection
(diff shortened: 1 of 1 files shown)


pyromaniac:
Could I ask you to add specs there?

syndbg:
@pyromaniac Sure, thing! I'll add some later today  (EST Time here) :)

pyromaniac:
Also how would it act if preload is enabled?

syndbg:
When the preload option is true, it would invoke a new method that does the
object loading and @objects assignment. #objects would return @objects if
not nil or invoke that method to assign @objects and then return them.

P.S writing from the phone. But be assured that I will refractor to take
into account the scenario described by you :)

pyromaniac:
Hey! Any progress here?

pyromaniac:
This feature was implemented in the new request DSL.

syndbg:
@pyromaniac Thank you. I was too long inactive on this for which I'm sorry.

PR #256, opened 2015-09-22
Bump elasticsearch version

gmile:
Few tests were failing for me on current master since I'm using 1.7.2 on my OS X. I understand this may not be the case for other users / developers, so please close or merge this at will.


pyromaniac:
Yeah, it is ok, thanks!

PR #257, opened 2015-09-22
Fix rspec-specific warnings

gmile:
This fixes a few rspec-related warnings I've spotted during initial test runs.

### DIFF
--- a/spec/chewy/index/actions_spec.rb
+++ b/spec/chewy/index/actions_spec.rb
@@ -54,8 +54,8 @@ describe Chewy::Index::Actions do
 
     context do
       before { DummiesIndex.create }
-      specify { expect { DummiesIndex.create! }.to raise_error }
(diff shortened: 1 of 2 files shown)


spec/chewy/index/actions_spec.rb
```diff
@@ -54,8 +54,8 @@
 
     context do
       before { DummiesIndex.create }
-      specify { expect { DummiesIndex.create! }.to raise_error }
-      specify { expect { DummiesIndex.create!('2013') }.to raise_error }
+      specify { expect { DummiesIndex.create! }.to raise_error(Elasticsearch::Transport::Transport::Errors::BadRequest) }
```
gmile: Hmm, I think the errors should be more descriptive. I will add message expectations also.

pyromaniac:
Thanks!

PR #264, opened 2015-10-07
Do not perform sidekiq workers if no ids present

AndreySavelyev:
When I set request_strategy to sidekiq I have many object updates. I have many sidekiq jobs for update in queue. It's ok. But if nil return in observe callback like this 
`update_index('users#user') { self if user_have_changes? }` I get many sidekiq jobs with `[]` argument

image

Maybe it would be better not to perform sidekiq worker when there no ids?

### DIFF
--- a/lib/chewy/strategy/sidekiq.rb
+++ b/lib/chewy/strategy/sidekiq.rb
@@ -19,7 +19,9 @@ module Chewy
       end
 
       def leave
-        @stash.all? { |type, ids| Chewy::Strategy::Sidekiq::Worker.perform_async(type.name, ids) }
+        @stash.each do |type, ids|
+          Chewy::Strategy::Sidekiq::Worker.perform_async(type.name, ids) unless ids.empty?
+        end
       end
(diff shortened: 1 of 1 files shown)


lib/chewy/strategy/sidekiq.rb
```diff
@@ -19,7 +19,11 @@ def perform(type, ids)
       end
 
       def leave
-        @stash.all? { |type, ids| Chewy::Strategy::Sidekiq::Worker.perform_async(type.name, ids) }
+        @stash.each do |type, ids|
+          if ids.any?
```
pyromaniac: Son't use `any?` use `empty?` instead

pyromaniac:
Great idea, actually

lib/chewy/strategy/sidekiq.rb
AndreySavelyev: fxd

pyromaniac:
Great work, thanks!

PR #266, opened 2015-10-09
ActiveJob strategy

mkcode:
I recently upgraded my Rails app to use ActiveJob and thought chewy could use it too. 

This works for me in my application. Tests pass for me as well. Was not sure about how or if this needs any additional Appraisal setup.

Cheers!

PR #267, opened 2015-10-11
Active job queue

mkcode:
Follow up to #266.

Uses the 'chewy' queue for active job jobs. Gets rid of excess active job logging in tests.


pyromaniac:
Thanks!

PR #277, opened 2015-11-04
use the configured assets prefix instead of hardcoding /assets

davekaro:
In our app we have configured a different assets prefix since we have an `AssetsController` with corresponding routes. In this controller, the request strategy is not being used, but should be.

### DIFF
--- a/lib/chewy/railtie.rb
+++ b/lib/chewy/railtie.rb
@@ -10,7 +10,7 @@ module Chewy
       end
 
       def call(env)
-        if env['PATH_INFO'].start_with?('/assets/')
+        if env['PATH_INFO'].start_with?(Rails.application.config.assets.prefix)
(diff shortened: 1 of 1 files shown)


pyromaniac:
Probably it make sense to suppress it optional

davekaro:
I'm not sure what you mean. Do you mean optionally suppress the strategy block?

I noticed https://github.com/toptal/chewy/commit/f0bdd209cd199505e4e083310a8cd5741224fe76 says it fixes something by avoiding wrapping assets in a strategy block, but I'm not clear on why that would break.

pyromaniac:
Never mind, I've just mixed things. I've thought at first that you need chewy strategy for /assets as well. And only the I realized :)

pyromaniac:
Thanks!

PR #280, opened 2015-11-13
Safely unsubscribe Notifications subscriber.

marshall-lee:
Hello!

Unsubscription should be done safely and here is the fix.

This bug is not as inoffensive as it seems at first sight. Block passed to `subscribe` *captures* method `*args` (fat activerecord objects, maybe with references to loaded associations). These captured objects are never garbage collected.

Needless to say that it also overwhelms the callback chain that slows everything down.

### DIFF
--- a/lib/chewy/type/import.rb
+++ b/lib/chewy/type/import.rb
@@ -57,9 +57,10 @@ module Chewy
             errors = args.last[:errors]
           end
           import *args
-          ActiveSupport::Notifications.unsubscribe(subscriber)
           raise Chewy::ImportFailed.new(self, errors) if errors.present?
           true
+        ensure
+          ActiveSupport::Notifications.unsubscribe(subscriber) if subscriber
         end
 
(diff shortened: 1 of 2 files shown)


marshall-lee:
The interesting fact here is that intuition says us that this block shouldn't capture method `*args` because it has its own shadowed `*args`.

But look at this example:

```ruby
class X
end

def m(x)
  proc do |x|
  end
end

a = m(X.new)

puts ObjectSpace.each_object(X).count
GC.start
puts ObjectSpace.each_object(X).count
```
This one prints:
```
1
1
```
To get rid of captured `x` argument we should do this thing:
```ruby
class X
end

def m(x)
  x = nil
  proc do |x|
  end
end

a = m(X.new)

puts ObjectSpace.each_object(X).count
GC.start
puts ObjectSpace.each_object(X).count
```
And we get:
```
1
0
```

A closure is a double edged sword :)

pyromaniac:
Awesome catch, thanks!

PR #284, opened 2015-11-18
Check current rspec.mock_framework instead of simple usage of defined?

mainameiz:
There is a bug when we have mocha in Gemfile but not use it as mock framework in rspec.


pyromaniac:
This actually makes sense, thanks!

mainameiz:
So fast :+1:

pyromaniac:
I've got notifications lol :)

PR #294, opened 2015-12-08
use .all when loading sequel models, identify is the same as in Orm, …

mrbrdo:
some improvements for the sequel adapter


pyromaniac:
Perfect, thanks!

jirutka:
This improvement has broken my code that works with 0.8.2 and 0.8.3…

PR #297, opened 2015-12-10
Add support for elasticsearch v2.0...v2.1

sergeygaychuk:
Tested on:
1. v1.7.3
2. v2.0
3. v2.1
    a) with delete-by-query plugin
    b) without delete-by-query plugin

Maybe someone can setup v2.x version on Travis CI.

### DIFF
--- a/lib/chewy/query.rb
+++ b/lib/chewy/query.rb
@@ -353,6 +353,7 @@ module Chewy
     # Returns empty hash if no facets was requested or resulted.
     #
     def facets params = nil
+      raise RemovedFeature, 'removed in elasticsearch 2.0' if Runtime.version >= '2.0'
       if params
         chain { criteria.update_facets params }
       else
--- a/lib/chewy/query.rb
+++ b/lib/chewy/query.rb
@@ -914,6 +914,12 @@ module Chewy
         type: _types.one? ? _types.first : _types do
           Chewy.client.delete_by_query(request)
       end
+    rescue Elasticsearch::Transport::Transport::Errors::BadRequest => e
+      if e.message =~ /No handler found for uri/
+        raise PluginMissing, "install delete-by-query plugin"
+      else
+        raise e
+      end
     end
 
     # Find all records matching a query.
--- a/lib/chewy/query.rb
+++ b/lib/chewy/query.rb
@@ -992,7 +992,7 @@ module Chewy
           begin
             Chewy.client.search(_request)
           rescue Elasticsearch::Transport::Transport::Errors::NotFound => e
-            raise e if e.message !~ /IndexMissingException/
+            raise e if e.message !~ /IndexMissingException/ && e.message !~ /index_not_found_exception/
             {}
           end
       end
--- a/lib/chewy/query.rb
+++ b/lib/chewy/query.rb
@@ -907,18 +907,16 @@ module Chewy
     #   UsersIndex::User.filter{ age <= 42 }.delete_all
     #
     def delete_all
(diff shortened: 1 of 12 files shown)


spec/chewy/fields/time_fields_spec.rb
```diff
@@ -21,7 +21,7 @@
   let(:range) { (time - 1.minute)..(time + 1.minute) }
 
   specify { expect(PostsIndex.total).to eq(3) }
-  specify { expect(PostsIndex.filter { published_at == o{range} }.count).to eq(2) }
-  specify { expect(PostsIndex.filter { published_at == o{range.min.utc..range.max.utc} }.count).to eq(2) }
-  specify { expect(PostsIndex.filter { published_at == o{[range.min.to_date..range.max.to_date]} }.count).to eq(1) }
```
sergeygaychuk: I have some troubles with that line on v2.x. I'll open issue on elastic.

pyromaniac:
Wow wow!

lib/chewy/query.rb
```diff
@@ -992,7 +997,7 @@ def _response
           begin
             Chewy.client.search(_request)
           rescue Elasticsearch::Transport::Transport::Errors::NotFound => e
-            raise e if e.message !~ /IndexMissingException/
+            raise e if e.message !~ /IndexMissingException/ && e.message !~ /index_not_found_exception/
```
pyromaniac: Which exceptions can be here as well?

sergeygaychuk: If you look at [this PR](https://github.com/elastic/elasticsearch/pull/12023/files), you will see that IndexMissingException was changed to IndexNotFoundException.

pyromaniac: Then should not it be || here to support both of versions?

sergeygaychuk: It's !~ (not match).

For example: 
[3] pry(main)> '{"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index","resource.type":"index_or_alias","resource.id":"products","index":"products"}],"type":"index_not_found_exception","reason":"no such index","resource.type":"index_or_alias","resource.id":"products","index":"products"},"status":404}'.!~(/IndexMissingException/)
=> true

'{"error":{"root_cause":[{"type":"index_not_found_exception","reason":"no such index","resource.type":"index_or_alias","resource.id":"products","index":"products"}],"type":"index_not_found_exception","reason":"no such index","resource.type":"index_or_alias","resource.id":"products","index":"products"},"status":404}'.!~(/index_not_found_exception/)
=> false

The same for strings which contains IndexMissingException

If we want to reraise exception, when string does not contain IndexMissingException and index_not_found_exception substrings, then we'll use next logic:
reraise exception if message doesn't contain IndexMissingException and index_not_found_exception.

pyromaniac: Ah you are right, sorry, I've mixed something

sergeygaychuk: Don't worry, I recheck it few times today :smile: Thanks

pyromaniac:
So are you considering this PR as ready to be merged?

sergeygaychuk:
I think yes. If I find some new issues with v2.x, then I'll do separate PR.

sergeygaychuk:
@pyromaniac But we need to add some tests to Travis CI. Really,  I don't  how to do it :(

pyromaniac:
I know :) Will do it eventually

sergeygaychuk:
:) Thanks

PR #298, opened 2015-12-10
Add note about Chewy.request_strategy in README

Tonkpils:
Without searching through the code, it wasn't very clear that one should use `request_strategy` to modify the controller strategy. Adding it to the README reduces the need to go through the code.


davidguthu:
I got bit by this.. +1

pyromaniac:
Thanks!

Issue #296, opened 2015-12-09
Pass a block to Elasticsearch::Client


davekaro:
I'd like to pass a block to configure the faraday instance of the Elasticsearch::Client. From [elasticsearch-transport readme](https://github.com/elastic/elasticsearch-ruby/tree/master/elasticsearch-transport#transport-implementations):

You can use any standard Faraday middleware and plugins in the configuration block, for example sign the requests for the AWS Elasticsearch service:

```
require 'patron'
require 'faraday_middleware/aws_signers_v4'

client = Elasticsearch::Client.new url: 'https://search-my-cluster-abc123....es.amazonaws.com' do |f|
  f.request :aws_signers_v4,
            credentials: Aws::Credentials.new(ENV['AWS_ACCESS_KEY'], ENV['AWS_SECRET_ACCESS_KEY']),
            service_name: 'es',
            region: 'us-east-1'
end
```

I believe https://github.com/toptal/chewy/blob/master/lib/chewy.rb#L125 is where the client is created, but I'm not sure the best way to pass a block to that. Could I simply override the client myself by setting `Thread.current[:chewy_client]` to whatever I want? Are there multi-threaded concerns I should be aware of if I do that?

pyromaniac:
Well, I would rather redefine the whole method in my application, also I would (if I would be on your place) create a patch since this is a good observation :)

PR #301, opened 2015-12-17
add client_class option

feymartynov:
A use case for it is setting up a connection to AWS Elasticsearch.

According to elasticsearch-transport's [readme](https://github.com/elastic/elasticsearch-ruby/tree/master/elasticsearch-transport#transport-implementations) one should pass a block to `::ElasticsearchClient#initialize` which adds necessary authorization headers to each request. However this can't be done with chewy.

A possible hack as it was discussed in the issue https://github.com/toptal/chewy/issues/296 is to assign `Thread.current[:chewy_client]` somewhere (say, in Rails initializer) with a custom instance. But this would affect only the current thread.

With `:client_class` option it becomes possible to perfom dependency injection of a custom client class which gets instantiated in every thread you call `Chewy.client`.

So an example initializer for AWS tuning looks like this:
```ruby
require 'faraday_middleware/aws_signers_v4'

class AwsElasticsearchClient < ::Elasticsearch::Client
  def initialize(options)
    super(options) do |f|
      f.request :aws_signers_v4,
                service_name: 'es',
                region: 'us-east-1',
                credentials: Aws::Credentials.new(
                  ENV['AWS_ACCESS_KEY'],
                  ENV['AWS_SECRET_ACCESS_KEY'])
    end
  end
end

Chewy.configuration[:chewy_client] = AwsElasticsearchClient
```

### DIFF
--- a/lib/chewy.rb
+++ b/lib/chewy.rb
@@ -122,7 +122,10 @@ module Chewy
     # Main elasticsearch-ruby client instance
     #
     def client
-      Thread.current[:chewy_client] ||= ::Elasticsearch::Client.new configuration
+      Thread.current[:chewy_client] ||= begin
+        klass = configuration[:client_class] || ::Elasticsearch::Client
+        klass.new configuration
+      end
     end
 
     # Sends wait_for_status request to ElasticSearch with status
--- a/spec/chewy_spec.rb
+++ b/spec/chewy_spec.rb
@@ -110,4 +110,18 @@ describe Chewy do
     specify { expect(DevelopersIndex.exists?).to eq(false) }
     specify { expect(CompaniesIndex.exists?).to eq(false) }
   end
+
+  describe '.client' do
+    let!(:initial_client) { Thread.current[:chewy_client] }
+    let(:klass) { Class.new(::Elasticsearch::Transport::Client) }
+
+    before do
+      Thread.current[:chewy_client] = nil
+      allow(Chewy).to receive_messages(configuration: { client_class: klass })
+    end
+
+    its(:client) { is_expected.to be_a klass }
+
+    after { Thread.current[:chewy_client] = initial_client }
+  end
 end


davekaro:
Awesome great idea. This would work perfect for #296

pyromaniac:
Hm, I was mostly thinking about transport_config settings proc, which is passed on transport initializaton.

feymartynov:
Created a new pull request for this: https://github.com/toptal/chewy/pull/302.

morgler:
Thanks for the pull request! But I have trouble using it correctly :(. I do this in my chewy.rb initializer:
 
```
  Chewy.settings = {
    host: 'http://my-es-instance-on-aws.us-east-1.es.amazonaws.com'
  }
  Chewy.configuration[:transport_options] = {
    headers: { content_type: 'application/json' },
    proc: -> (f) do
        Rails.logger.debug ">>> Calling proc for chewy client"
        f.request :aws_signers_v4,
                  service_name: 'es',
                  region: 'us-east-1',
                  credentials: Aws::Credentials.new(
                    Rails.application.secrets.aws_access_key_id,
                    Rails.application.secrets.aws_secret_access_key)
    end
  }
```

If I try to use chewy afterwards, I always get a `Elasticsearch::Transport::Transport::Errors::Forbidden` error as if I would not have supplied the AWS credentials. I inserted a log line in the proc and it is never reached.

I also deleted chewy.yml to avoid collisions of it with the initializer.

What am I doing wrong? Maybe you could add an example to the readme of how to use this proc correctly.

morgler:
Okay, rubber ducking helped ;). It seems you can't mix the configuration methods. So I had to write it all in one block:

```
  Chewy.settings = {
    host: 'http://my-es-instance-on-aws.us-east-1.es.amazonaws.com',
    transport_options: {
      headers: { content_type: 'application/json' },
      proc: -> (f) do
          Rails.logger.info ">>> Calling proc for chewy client"
          f.request :aws_signers_v4,
                    service_name: 'es',
                    region: 'us-east-1',
                    credentials: Aws::Credentials.new(
                      Rails.application.secrets.aws_access_key_id,
                      Rails.application.secrets.aws_secret_access_key)
      end
    }
  }
```

Issue #271, opened 2015-10-19
Importing to amazon elastic search service


drnluz:
Hello. I'm trying to import a model to the amazon elasticsearch service but I'm getting the following error:

`Elasticsearch::Transport::Transport::Errors::Forbidden: [403] {"message":"When Content-Type:application/x-www-form-urlencoded, URL cannot include query-string parameters (after '?'): '/all/factory/_bulk?refresh=false'"}`

Digging a little I found that amazon ES service does not accept query parameters.
Is there a way to run `MyIndex.reset!` without sending `refresh` as a query parameter?

pyromaniac:
Well, it is quite hardcoded here: https://github.com/toptal/chewy/blob/ef0fab43cbd302997fcc2b2ae3b400ff97687f4c/lib/chewy/type/import.rb#L22
But you may send a patch of de-hardcoding :) Like, in case of `refresh: false` don't add anything.

pyromaniac:
I didn't expect such a behavior, Amazon engineers are strange guys, ES itself doesn't have such a strict rules

yurifrl:
Hy @drnluz , any updates on this issue?

feymartynov:
Had the same issue with AWS. Fixed with explicitly specifying Content-Type header in `chewy.yml`:
```yaml
production:
  host: 'some-host.es.amazonaws.com:80'
  transport_options:
    headers:
      content_type: 'application/json'
```

pyromaniac:
Perfect, thanks for the solution!

PR #302, opened 2015-12-21
add proc transport option

feymartynov:
A use case for it is setting up a connection to AWS Elasticsearch.

According to elasticsearch-transport's readme one should pass a block to ::ElasticsearchClient#initialize which adds necessary authorization headers to each request. However this can't be done with chewy.

A possible hack as it was discussed in the issue #296 is to assign Thread.current[:chewy_client] somewhere (say, in Rails initializer) with a custom instance. But this would affect only the current thread.

With `:proc` option it becomes possible to pass a block to `Elasticsearch::Constructor`.

So an example initializer for AWS tuning looks like this:

```ruby
require 'faraday_middleware/aws_signers_v4'

Chewy.configuration[:transport_options] = {
  headers: { content_type: 'application/json' },
  proc: -> (f) do
      f.request :aws_signers_v4,
                service_name: 'es',
                region: 'us-east-1',
                credentials: Aws::Credentials.new(
                  ENV['AWS_ACCESS_KEY'],
                  ENV['AWS_SECRET_ACCESS_KEY'])
  end
}
```

P.S. For the purpose of `Content-Type` header see issue https://github.com/toptal/chewy/issues/271.

### DIFF
--- a/lib/chewy.rb
+++ b/lib/chewy.rb
@@ -122,7 +122,10 @@ module Chewy
     # Main elasticsearch-ruby client instance
     #
     def client
-      Thread.current[:chewy_client] ||= ::Elasticsearch::Client.new configuration
+      Thread.current[:chewy_client] ||= begin
+        block = configuration[:transport_options].try { |c| c[:proc] }
+        ::Elasticsearch::Client.new(configuration, &block)
+      end
     end
 
     # Sends wait_for_status request to ElasticSearch with status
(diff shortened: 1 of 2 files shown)


pyromaniac:
Thanks!

rromanchuk:
o_O  am i on an incorrect version? or is there something fundamental i'm not understanding here?
screen shot 2017-03-10 at 6 48 17 pm

PR #303, opened 2015-12-26
Guard against assets not being configured

clupprich:
The change in #277 unfortunately breaks Rails applications in API mode (Rails 5 or via the `rails-api` gem): When configuring an application API only, the `Rails.application.config.assets` entry isn't available and the `RequestStrategy` middleware crashes the application. This now checks if the application is running on api-only mode and does the test for the assets-path afterwards.

### DIFF
--- a/lib/chewy/railtie.rb
+++ b/lib/chewy/railtie.rb
@@ -10,7 +10,8 @@ module Chewy
       end
 
       def call(env)
-        if env['PATH_INFO'].start_with?(Rails.application.config.assets.prefix)
+        # For Rails applications in `api_only` mode, the `assets` config isn't present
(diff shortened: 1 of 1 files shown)


pyromaniac:
Why don't check `Rails.application.config.assets` entry itself here? It will be more clear I suppose

clupprich:
I tried that, but that's hidden behind some method_missing magic.

clupprich:
Also tried the usual suspects like #key?, and checked the documentation, but seems that you can't query if a config entry exists. We could try a respond_to, however.

clupprich:
@pyromaniac I've changed the code now to check if the `assets` entry is there via `respond_to?`.

pyromaniac:
Perfect, thanks!

clupprich:
Awesome!

PR #304, opened 2016-01-02
Simplify Chewy::Type::Import#bulk_body a little

DNNX:
http://ruby-doc.org/core-2.3.0/Enumerable.html#method-i-flat_map


pyromaniac:
Great, thanks!

PR #306, opened 2016-01-05
Better error handling in Chewy::Strategy#wrap

barthez:
Passing non-exising strategy to Chewy::Strateg#wrap caused unclear error: "Can't pop root strategy". Now, push and wrap give back better error when strategy is missing.

### DIFF
--- a/lib/chewy/strategy.rb
+++ b/lib/chewy/strategy.rb
@@ -56,10 +56,10 @@ module Chewy
     end
 
     def wrap name
-      push name
+      stack = push(name)
       yield
     ensure
-      pop
+      pop if stack
     end
 
   private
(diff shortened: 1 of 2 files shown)


barthez:
It turns out that with mongoid we can observe strange behavior of `safe_constantize` (`const_get` in particular). It turns you that `Foo.const_get('Bar')` raises error with message `uninitialized constant Bar` (what cannot be handled by safe_constantize from ActiveSupport < 4.2), but `Foo.const_get(:Bar)` raises error with message `uninitialized constant Foo::Bar` and further calls of `const_get` with string will raise "proper" error message, with full qualified constant.


I've introduced small [workaround](https://github.com/toptal/chewy/pull/306/files#diff-d86521f4fc4f30558ca8d7778404c304R78) which handle that case.

pyromaniac:
Perfect, thanks!

Issue #225, opened 2015-07-30
Elasticsearch, Chewy, Postgres Schemas, Multi-Tenancy


eliduke:
I have recently switched from MySQL over to Postgres with multiple schemas, and am having trouble with `rake chewy:reset:all`. It looks as though it is defaulting to the "public" schema, but I want to specify the schema. I am using the apartment gem, so I put this in one of my indexes:

`Apartment::Tenant.switch!('tenant_name')`

And that fixed the problem, but that got me thinking bigger about elasticsearch and multi-tenancy in general. Does chewy have any sort of implementation of that? If not, do you have any recommendations?

pyromaniac:
Chewy does nothing with schemas, it uses queries provided by you. And query already should know which tenant it should use.

eliduke:
Here's the breakdown:

My app knows what tenant because it's one of the necessary headers needed for authentication with the API. So, within the app it is all the correct tenant. 

When going into the rails console, we have a little bit of config code that checks for all available tenants and asks which one you want before fully loading the console. That's awesome.

When I console into our main tenant (not the standard "public" tenant), I have all the expected records in the database. But when I run the "rake chewy:reset:all", it indexes the data in the "public" scheme, instead of the scheme that **the rest of the app** is using. Why is that?

It seems like y'all need to add some sort of tenant config option where I can tell chewy to index the current tenant instead of the public one. Maybe it is just a config option specifying MySQL or Postgres or what-have-you, with specific adapter type stuff for the chosen database type.

Your thoughts?

pyromaniac:
Actually you have to prepare environment for the `rake chewy:reset` task as well as for every other task. You may create a preparation task like `setup_tenant` and use it manually like `rake setup_tenant[name] chewy:reset` or even prepend this task with rake DSL to every rake task automatically. This will make rake tasks behavior similar to console. Chewy should know nothing about your tenants.

yurifrl:
Can i sugest a change to index_name for something like:
````ruby
def self.index_name(suggest = nil)
    prefix = Chewy.configuration[:prefix]

    prefix = prefix.call if prefix.respond_to? :call

    if suggest
      @index_name = build_index_name(suggest, prefix: prefix)
    else
      @index_name ||= begin
        build_index_name(
          name.sub(/Index\Z/, '').demodulize.underscore,
          prefix: prefix
        ) if name
      end
    end
    @index_name or raise UndefinedIndex
  end
````
so it's possible to do something like:
````ruby
[...]

PR #314, opened 2016-01-20
Refactor/extract method on prefix

mikeyhogarth:
There's an open issue https://github.com/toptal/chewy/issues/225 where a monkey patch is suggested in order to get Chewy working nicely in multitenant environments.

This change will make that monkey patch much smaller to implement (you would only need to override the `default_prefix` method) as well as generally reducing repetition in the code (less references to the same configuration option).

It also paves the way for more advanced prefix-calculating functionality in the future.

### DIFF
--- a/lib/chewy/index.rb
+++ b/lib/chewy/index.rb
@@ -30,12 +30,12 @@ module Chewy
     #
     def self.index_name(suggest = nil)
       if suggest
-        @index_name = build_index_name(suggest, prefix: prefix)
+        @index_name = build_index_name(suggest, prefix: default_prefix)
       else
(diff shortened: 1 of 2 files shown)


pyromaniac:
Is it ready to be merged?

mikeyhogarth:
Absolutely.

mikeyhogarth:
I mean, there will be a second PR incoming with an additional change but that one needs a bit more discussion than this does so I figured I'd do it separately.  Cheers.

pyromaniac:
Perfect, thanks!

mikeyhogarth:
Cheers!

PR #318, opened 2016-01-23
Remove memoization from index name

mikeyhogarth:
Simple change but maybe one that needs some discussion. The reason for making this change is that the value that needs to be assigned to `@index_name` may change between calls to `index_name`, such as between requests in multitenant environment. 

If for example the `default_prefix` were set to a lambda or were overridden to be something dynamic, it wouldn't be re-initialised between requests.

We can sort this out by just not memoizing that variable, however this does introduce a performance overhead (it'll re-evaluate the regex/demodulize etc. every time and will create a new variable) however I think this overhead is very minor compared to the benefit. 

If you'd prefer not to have the overhead at all I could introduce this as a configurable option (`dynamic_index_names: true` or something) but then that's maybe a step too far.

### DIFF
--- a/lib/chewy/index.rb
+++ b/lib/chewy/index.rb
@@ -32,7 +32,7 @@ module Chewy
       if suggest
         @index_name = build_index_name(suggest, prefix: default_prefix)
       else
-        @index_name ||= begin
+        @index_name = begin
           build_index_name(
             name.sub(/Index\Z/, '').demodulize.underscore,
             prefix: default_prefix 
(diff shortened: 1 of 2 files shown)


mikeyhogarth:
Hmm this appears to cause the test suite to fail :/

@pyromaniac do you know if the memoization here (not being able to change the non-suggested index name after it's been set once) is desired functionality? The fact that I've broke the test suite by removing memoization would suggest so

pyromaniac:
Well, if we are going to remove memoization, we have to get rid of instance variables completely. Also we have to extract `name.sub(/Index\Z/, '').demodulize.underscore` and similar statements to memoize them separately.

pyromaniac:
But I'm ok with memoization removement

pyromaniac:
Hey man! Are you going to update this? I would like to merge thi PR

pyromaniac:
In master already

PR #319, opened 2016-01-24
Fixed base field compose when hash_key is a symbol and hash_value is …

eproulx-petalmd:
An object field whose value is an hash with symbol keys and one of them contains a `false` value. The corresponding field will be indexed with `null` instead of `false`

example :
```ruby
field :person, value: (task, crutch) -> {crutch.person[task.id]} do
  field :name, type:string
  field :is_male, type: boolean
end
```
Let's say that `crutch.person[task.id]` would return 
```ruby
{
  name: 'Jane Doe'
  is_male: false
}
```

`is_male` would be index has `null` instead of `false`

### DIFF
--- a/lib/chewy/fields/base.rb
+++ b/lib/chewy/fields/base.rb
@@ -39,10 +39,16 @@ module Chewy
             value.call(*objects.first(value.arity))
           end
         elsif object.is_a?(Hash)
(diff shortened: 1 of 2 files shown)


pyromaniac:
Perfect, thanks!

PR #324, opened 2016-02-03
Type#method_missing calls super

jesjos:
Earlier implementation returned nil on unknown method calls, which would make debugging much harder.


pyromaniac:
:+1:

pyromaniac:
Thanks!

PR #335, opened 2016-02-20
Use scoped subscription to ActiveSupport::Notifications

0x0badc0de:
When chewy tasks are invoked multiple times from the same environment - their stats are printed as many times as they were invoked. Easiest way to see it is with `rake chewy:update[model1] chewy:reset[model2]`.
This PR makes subscriptions scoped to the task block only.


pyromaniac:
Thanks!

PR #336, opened 2016-02-23
Remove redundant preloading from the example describing Crutches™ Technology

DNNX:
I might be wrong here, but it seems that it is not really necessary because we use Advanced Crutches™ Technology to load category names. Please correct me if I'm missing something.


pyromaniac:
Thanks!

PR #337, opened 2016-02-24
Use timestamp column to order batches while importing

barthez:
Importing using ascending identifiers for order have some cons, like item updated in different source (e.g. webserver) while it was already imported to newly created index (while resetting) will be outdated when we switch the index aliases.                                                       
                                                                         
To prevent this behavior we can check if it possible to use timestamp column for sorting. Fetching batches sorted by e.g update_at column will prevent described behavior, but will make import a bit longer.           
                                                                         
As far as I can see it will never cause infinite loops (need to be properly tested with specs).                                                      
                                                                         
For now it includes only support for ActiveRecord, but seems that adding Mongoid and Sequel support is a manner of extracting code to Orm adapter (with small tunning).

### DIFF
--- a/lib/chewy/config.rb
+++ b/lib/chewy/config.rb
@@ -36,7 +36,19 @@ module Chewy
       # in tests with transactional fixtures or transactional
       # DatabaseCleaner strategy.
       #
-      :use_after_commit_callbacks
+      :use_after_commit_callbacks,
+
+      # Use timestamp (updated_at, updated_on) ordering for import.
+      # False by default (order by id). It makes imports last
+      # a bit longer, but it also ensures that any items created or
+      # updated during import will be indexed properly. Only
+      # ActiveRecord adapter is supported.
+      #
+      # WARNING: In rare cases in could cause infinite imports,
+      # read Chewy:Type:Adapter:ActiveRecord:InfiniteImportWarning
+      # message for details.
+      #
+      :use_timestamp_ordered_import
 
     def self.delegated
       public_instance_methods - self.superclass.public_instance_methods - Singleton.public_instance_methods
@@ -49,6 +61,7 @@ module Chewy
       @root_strategy = :base
       @request_strategy = :atomic
       @use_after_commit_callbacks = true
+      @use_timestamp_ordered_import = false
     end
 
     def transport_logger= logger
--- a/lib/chewy/index.rb
+++ b/lib/chewy/index.rb
@@ -152,14 +152,6 @@ module Chewy
       public_methods - Chewy::Index.public_methods - type_names.map(&:to_sym)
     end
 
-    def self.timestamp_ordered_import(flag = nil)
-      if flag.nil?
-        @timestamp_ordered_import == true
-      else
-        @timestamp_ordered_import = flag
-      end
-    end
-
   private
 
     def self.build_index_name *args
--- a/lib/chewy/index.rb
+++ b/lib/chewy/index.rb
@@ -152,6 +152,14 @@ module Chewy
       public_methods - Chewy::Index.public_methods - type_names.map(&:to_sym)
     end
 
+    def self.timestamp_ordered_import(flag = nil)
+      if flag.nil?
+        @timestamp_ordered_import == true
+      else
+        @timestamp_ordered_import = flag
+      end
+    end
+
   private
 
     def self.build_index_name *args
--- a/lib/chewy/type/adapter/active_record.rb
+++ b/lib/chewy/type/adapter/active_record.rb
@@ -4,6 +4,28 @@ module Chewy
   class Type
     module Adapter
       class ActiveRecord < Orm
+        IMPORT_OVERHEAD_LIMIT = 2
+        class InfiniteImportWarning < RuntimeError
+          def initialize(batch_size)
+            @batch_size = batch_size
+          end
+
+          def message
+            <<-MSG.strip_heredoc
+              Import process indexed already #{IMPORT_OVERHEAD_LIMIT}x
+              as many object as there were in database when import
+              started. It probably means that you are creating or
+              updating more objects that are indexed in one batch
+              during time batch get processed.
+
+              Possible solutions:
+               - Increase batch size (current size: #{@batch_size})
+               - Boost batch import by making index entry generation faster
+                 (e.g. check if your type definition do not have N+1 query
+                 problem)
+            MSG
+          end
+        end
 
         def self.accepts?(target)
           defined?(::ActiveRecord::Base) && (
@@ -33,6 +55,9 @@ module Chewy
         def import_scope_ordered_by_id(scope, batch_size)
           result = true
 
+          base_number_of_items = scope.count
+          current_number_of_items = 0
+
           scope = scope.reorder(id_column.asc).limit(batch_size)
 
           ids = pluck_ids(scope)
@@ -40,6 +65,10 @@ module Chewy
           while ids.present?
             result &= yield grouped_objects(default_scope_where_ids_in(ids))
             break if ids.size < batch_size
+
+            current_number_of_items += ids.size
+            raise InfiniteImportWarning, batch_size if current_number_of_items > IMPORT_OVERHEAD_LIMIT*base_number_of_items
+
             ids = pluck_ids(scope.where(id_column.gt(ids.last)))
           end
 
@@ -49,6 +78,9 @@ module Chewy
         def import_scope_ordered_by_timestamp(scope, batch_size)
           result = true
 
+          base_number_of_items = scope.count
+          current_number_of_items = 0
+
           scope = scope.reorder(timestamp_column.asc, id_column.asc).limit(batch_size)
 
           ids_and_dates = pluck_ids_and_dates(scope)
@@ -56,6 +88,10 @@ module Chewy
           while ids_and_dates.present?
             result &= yield grouped_objects(scope_where_ids_in(scope, ids_and_dates.map(&:first)))
             break if ids_and_dates.size < batch_size
+
+            current_number_of_items += ids_and_dates.size
+            raise InfiniteImportWarning, batch_size if current_number_of_items > IMPORT_OVERHEAD_LIMIT*base_number_of_items
+
             last_id, last_updated_at = ids_and_dates.last
             ids_and_dates = pluck_ids_and_dates(scope.where(next_timestamp_ordered_batch_condition(last_id, last_updated_at)))
           end
--- a/lib/chewy/type/adapter/active_record.rb
+++ b/lib/chewy/type/adapter/active_record.rb
@@ -22,38 +22,46 @@ module Chewy
           @default_scope = @default_scope.reorder(nil).limit(nil).offset(nil)
         end
 
-        def import_scope(scope, batch_size, sort_by_updated_at)
+        def import_scope(scope, batch_size, sort_by_updated_at, &block)
+          if sort_by_updated_at && timestamp_column_name
+            import_scope_ordered_by_timestamp(scope, batch_size, &block)
+          else
+            import_scope_ordered_by_id(scope, batch_size, &block)
+          end
+        end
+
+        def import_scope_ordered_by_id(scope, batch_size)
           result = true
 
-          if !sort_by_updated_at || timestamp_column_name.nil?
-            scope = scope.reorder(id_column.asc).limit(batch_size)
+          scope = scope.reorder(id_column.asc).limit(batch_size)
 
-            ids = pluck_ids(scope)
+          ids = pluck_ids(scope)
 
-            while ids.present?
-              result &= yield grouped_objects(default_scope_where_ids_in(ids))
-              break if ids.size < batch_size
-              ids = pluck_ids(scope.where(id_column.gt(ids.last)))
-            end
-          else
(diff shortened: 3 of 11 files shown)


lib/chewy/type/adapter/active_record.rb
```diff
@@ -22,31 +22,65 @@ def cleanup_default_scope!
...
+        def timestamp_column
+          target.arel_table[timestamp_column_name]
+        end
+
         def pluck_ids(scope)
           scope.except(:includes).uniq.pluck(target.primary_key.to_sym)
         end
 
+        def pluck_ids_and_dates(scope)
+          scope.except(:includes).uniq.pluck(target.primary_key.to_sym, timestamp_column_name)
```
barthez: This is not supported in Rails 3.2, `pluck` accepts only one parameter.

DNNX: :(

lib/chewy/index.rb
```diff
@@ -152,6 +152,14 @@ def self.scopes
...
     end
 
+    def self.timestamp_ordered_import(flag = nil)
+      if flag.nil?
+        @timestamp_ordered_import == true
+      else
+        @timestamp_ordered_import = flag
+      end
+    end
+
```
DNNX: There is one problem with this: all inheritors of `Chewy::Index` will share the same `@timestamp_ordered_import` variable.

lib/chewy/type/adapter/active_record.rb
```diff
@@ -22,31 +22,65 @@ def cleanup_default_scope!
...
+          ids = pluck_ids(scope)
+
           while ids.present?
             result &= yield grouped_objects(default_scope_where_ids_in(ids))
             break if ids.size < batch_size
-            ids = pluck_ids(scope.where(target_id.gt(ids.last)))
+            ids = pluck_ids(scope.where(id_column.gt(ids.last)))
           end
 
           result
```
DNNX: After thinking about it for a while, I found a solution which might help to make the code a bit simpler and satisfy :cop:.

Didn't test it:

```ruby
def import_scope_ordered_by_timestamp(scope, batch_size)
  result = true

  # TODO: can the last 2 parameters be moved to TimestampOrderedBatches?
  TimestampOrderedBatches
    .new(scope, batch_size, id_column, timestamp_column)
    .each do |batch|
    result &= batch
  end

  result
end

class TimestampOrderedBatches
  def initialize(scope, batch_size, id_column, timestamp_column)
    @scope = scope
    @batch_size = batch_size
    @id_column = id_column
    @timestamp_column = timestamp_column
  end

  def each
    ids_and_dates = first_batch_ids_and_dates

    while ids_and_dates.present?
      yield grouped_objects(scope_where_ids_in(ordered_scope, ids_and_dates.map(&:first)))
      break if ids_and_dates.size < batch_size
      last_id, last_updated_at = ids_and_dates.last
      ids_and_dates = next_batch_ids_and_dates(last_id, last_updated_at)
    end
  end

  private

  attr_reader :scope, :batch_size, :id_column, :timestamp_column

  def first_batch_ids_and_dates
    pluck_ids_and_dates(first_batch_scope)
  end

  def next_batch_ids_and_dates(last_id, last_updated_at)
    pluck_ids_and_dates(next_batch_scope(last_id, last_updated_at))
  end

  def first_batch_scope
    ordered_scope
  end

  def next_batch_scope(last_id, last_updated_at)
    ordered_scope.where(next_timestamp_ordered_batch_condition(last_id, last_updated_at))
  end

  def pluck_ids_and_dates(scope)
    scope.except(:includes).uniq.pluck(id_column, timestamp_column)
  end

  def ordered_scope
    @ordered_scope ||= scope.reorder(timestamp_column_asc, id_column_asc).limit(batch_size)
  end
end
```

>

lib/chewy/type/adapter/active_record.rb
```diff
@@ -22,31 +22,65 @@ def cleanup_default_scope!
...
+
+          while ids_and_dates.present?
+            result &= yield grouped_objects(scope_where_ids_in(scope, ids_and_dates.map(&:first)))
+            break if ids_and_dates.size < batch_size
+            last_id, last_updated_at = ids_and_dates.last
+            ids_and_dates = pluck_ids_and_dates(scope.where(next_timestamp_ordered_batch_condition(last_id, last_updated_at)))
+          end
+
+          result
+        end
```
DNNX: that comment ^ should belong here

lib/chewy/type/adapter/active_record.rb
DNNX: There are 3 solutions that I can see now:

* Use `select(:id, :updated_at)`. That's bad performance-wise.
* Include the backport gem (`pluck_all` was the name, IIRC). Again, an extra dependency is not perfect.
* Roll out our own `pluck_all` by copying the code from somewhere and throwing out stuff that we don't need. Feels dirty.
* There is a way to return array of tules (or hashes) from the database. I've done it a couple of times. Don't remember the details, but it wasn't too hard. `AR::Base.select_all(scope)`, or something like this.

And I guess we need to check whether method `pluck` arity can be > 1 and fall back to our workaround only in case if it is not supported.

spec/chewy/type/adapter/active_record_spec.rb
```diff
@@ -63,7 +65,10 @@
   describe '#import' do
     def import(*args)
       result = []
-      subject.import(*args) { |data| result.push data }
+      subject.import(*args) do |data|
+        result.push data
+        yield if block_given?
+      end
       result
     end
```
DNNX: An alternative solution. A bit too fancy because it uses the fact that `each` returns the enumerable as array if block is given. If block is not given, it returns a lazy enumerable. This is why we need `.to_a` at the end.

```ruby
def import(*args, &block)
  subject.to_enum(:each, *args).each(&block).to_a
end
```

spec/chewy/type/adapter/active_record_spec.rb
```diff
@@ -104,6 +109,55 @@ def import(*args)
 
       specify { expect(import(cities.first, nil)).to eq([{index: [cities.first]}]) }
       specify { expect(import(cities.first.id, nil)).to eq([{index: [cities.first]}]) }
+
+      context 'TEST' do
```
DNNX: We need to review this spec also.

spec/chewy/type/adapter/active_record_spec.rb
barthez: Have you meant `subject.to_enum(:import, *args).each(&block).to_a`.

It is super-fancy, like lot of code here :wink:.

lib/chewy/index.rb
barthez: Afaik class instance variables are not shared between subclasses.

DNNX: You are right, ignore my comment.

spec/chewy/type/adapter/active_record_spec.rb
DNNX: Yea :import right

barthez: This is `Array#each` behavior to return self despite block is given or not. [Enumerator#each](http://ruby-doc.org/core-2.2.0/Enumerator.html#method-i-each) behaves differently when block is given.

DNNX: Good to know, thanks :)

barthez:
@pyromaniac What do you think about this?

lib/chewy/config.rb
```diff
@@ -36,7 +36,19 @@ class Config
...
+      :use_after_commit_callbacks,
+
+      # Use timestamp (updated_at, updated_on) ordering for import.
+      # False by default (order by id). It makes imports last
+      # a bit longer, but it also ensures that any items created or
+      # updated during import will be indexed properly. Only
+      # ActiveRecord adapter is supported.
+      #
+      # WARNING: In rare cases in could cause infinite imports,
+      # read Chewy:Type:Adapter:ActiveRecord:InfiniteImportWarning
```
DNNX: Probably it is better to use double colons here (`::`).

lib/chewy/type/adapter/active_record.rb
```diff
@@ -4,6 +4,44 @@ module Chewy
...
+            MSG
+          end
+        end
+
+        class InfiniteLoopIndicator
+          attr_reader :number
+
+          def initialize(base_number, limit_factor, batch_size)
+            @number = 0
+            @limit = limit_factor*base_number
```
DNNX: Is Rubocop ok with no spaces around `*`?

lib/chewy/type/adapter/active_record.rb
```diff
@@ -4,6 +4,44 @@ module Chewy
...
+              Possible solutions:
+               - Increase batch size (current size: #{@batch_size})
+               - Boost batch import by making index entry generation faster
+                 (e.g. check if your type definition do not have N+1 query
+                 problem)
+            MSG
+          end
+        end
+
+        class InfiniteLoopIndicator
```
DNNX: `InfiniteLoopDetector` maybe? I have no strong opinion on that, it just feels more accurate (to me).

lib/chewy/type/adapter/active_record.rb
```diff
@@ -22,31 +60,70 @@ def cleanup_default_scope!
...
         end
 
+        def pluck_ids_and_dates(scope)
+          return pluck_ids_and_dates_failover(scope) if multifield_pluck_missing?
+          scope.except(:includes).uniq.pluck(target.primary_key.to_sym, timestamp_column_name)
+        end
+
         def scope_where_ids_in(scope, ids)
-          scope.where(target_id.in(Array.wrap(ids)))
+          scope.where(id_column.in(Array.wrap(ids)))
```
DNNX: Is it possible for `ids` not to be an `Array`? Can the `wrap` call be removed?

lib/chewy/type/adapter/active_record.rb
```diff
@@ -56,6 +133,34 @@ def relation_class
...
+            .or(timestamp_column.eq(last_updated_at).and(id_column.gt(last_id)))
+        end
+
+        def multifield_pluck_missing?
+          Gem::Version.new(::ActiveRecord::VERSION::STRING) < Gem::Version.new('4.0')
+        end
+
+        def pluck_ids_and_dates_failover(scope)
+          scope = scope.except(:includes).uniq
+            .select([target.primary_key.to_sym, timestamp_column_name])
```
DNNX: If I remember things correctly, square brackets are not necessary here.

spec/chewy/type/adapter/active_record_spec.rb
```diff
@@ -104,6 +109,91 @@ def import(*args)
...
+
+      context 'timestamp ordered import' do
+        around do |example|
+          begin
+            previous_value = Chewy.use_timestamp_ordered_import
+            Chewy.use_timestamp_ordered_import = true
+            Timeout.timeout(1) { example.run }
+          ensure
+            Chewy.use_timestamp_ordered_import = previous_value
+          end
```
DNNX: Nice!

spec/chewy/type/adapter/active_record_spec.rb
```diff
@@ -104,6 +109,91 @@ def import(*args)
...
+            results = import(City.unscoped, batch_size: 2) do
+              cities.last.touch unless update_is_done
+              update_is_done = true
+            end
+
+            expect(results).to eq([{index: cities.last(2).reverse}, {index: [cities.first, cities.last]}])
+          end
+        end
+
+        context 'finish even objects are updated every batch' do
```
DNNX: _finishes_ ...
even _if/when_ objects ...

spec/chewy/type/adapter/active_record_spec.rb
```diff
@@ -104,6 +109,91 @@ def import(*args)
...
+
+            expect(results).to eq([
+              {index: cities.last(2).reverse},
+              {index: [cities.first, cities.last]},
+              {index: [cities[1]]}
+            ])
+          end
+        end
+
+        context 'fails if there is more objects updated every batch than batch size' do
```
DNNX: there _are_ more objects

spec/chewy/type/adapter/active_record_spec.rb
```diff
@@ -104,6 +109,91 @@ def import(*args)
...
+        context 'and timestamp is equal' do
+          let(:update_time) { Time.now }
+          let!(:cities) { Array.new(3) { City.create!(updated_at: update_time) } }
+          let!(:deleted) { nil }
+
+          specify { expect(import(City.unscoped, batch_size: 2))
+            .to eq([{index: cities.first(2)}, {index: cities.last(1)}]) }
+        end
+
+        context 'failback to id ordered import when field is missing' do
```
DNNX: _when **timestamp** field is missing_ is probably slightly more exact

DNNX:
Cool stuff. I'd like to hear what @pyromaniac thinks about it too :)

pyromaniac:
Well, it will be slower then ordering by id, we will need indexes on updated_at column. Also I'm not sure this solution will resolve all the our issues. Let's finish it and will see

lib/chewy/type/adapter/active_record.rb
barthez: It is called in AR 3.2. `select` accepts only 1 argument.

PR #339, opened 2016-02-26
Sanitize Elasticsearch::Client config

### LINKED ISSUE #333: specifying transport_options => proc give no method found error
I was trying to specify a custom Proc in the settings for adding faraday middleware.  Looking at the code here

https://github.com/toptal/chewy/blob/master/lib/chewy.rb#L126

it looks like you should be able to specify this in the settings and the proc will get passed to the client initialization code.  However, when I do this I get an error and I think it's the result of the config option finding it's way into the *configuration* variable.  I could probably produce a fix but i'm not familiar enough with the testing framework and development process to write an entire test case around it...

reidab:
This avoids passing invalid config values to Elasticsearch::Client.new by stripping `prefix` and `transport_options[proc]` from the config hash.

Fixes #333

### DIFF
--- a/lib/chewy.rb
+++ b/lib/chewy.rb
@@ -123,8 +123,10 @@ module Chewy
     #
     def client
       Thread.current[:chewy_client] ||= begin
-        block = configuration[:transport_options].try { |c| c[:proc] }
-        ::Elasticsearch::Client.new(configuration, &block)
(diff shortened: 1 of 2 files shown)


pyromaniac:
Perfect, thanks!

PR #340, opened 2016-02-26
Allow set import params for rake tasks

barthez:
In certain cases on batch import while resetting takes so much time (due slow bulk entry generation) that items from batch can be updated from a different source (like a web server). Tuning this parameter allows making single batch import fast enough that example situation never happen.

For a sake of symmetry, this change also allows you to set `refresh` param for import from a console.

@pyromaniac What do you think?


pyromaniac:
I just don't think it make sense to pass `refresh` as env variable, otherwise - great idea

barthez:
I've removed support of `CHEWY_REFRESH` env then.

pyromaniac:
Also maybe it will be even better to ass per-type config? Like, default batch size for every type

barthez:
Right, usually not all types are similarly complex. Will prepare separate PR with that improvement.

barthez:
@pyromaniac: Is everything all right now?

pyromaniac:
Are you sure we need this if we have configurable per-type options?

barthez:
Yeah... It will overwrite all per type setting what makes this thing useless. Closing...

PR #341, opened 2016-02-26
Add type specific settings

barthez:
As mentioned in https://github.com/toptal/chewy/pull/340 adding per type settings for import options would be a nice idea. It allows to tune batch size for each defined type specifically.

@pyromaniac What do you think about it?

### DIFF
--- a/lib/chewy/type.rb
+++ b/lib/chewy/type.rb
@@ -12,6 +12,8 @@ require 'chewy/type/adapter/sequel'
 
 module Chewy
   class Type
+    IMPORT_OPTIONS_KEYS = %i[batch_size]
+
     include Search
     include Mapping
     include Wrapper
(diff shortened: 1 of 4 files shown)


lib/chewy/type.rb
```diff
@@ -47,6 +51,10 @@ def self.scopes
       public_methods - Chewy::Type.public_methods
     end
 
+    def self.settings(params)
+      self._settings = self._settings.deep_merge(params)
```
barthez: should params keys be restricted to "known settings"? Like:
```
KNOWN_SETTINGS = %i[import_options]

...
def self.settings(params)
  unknown_settings = params.each_key.select { |key| !KNOWN_OPTIONS.include?(key) }
  raise ArgumentError, "Unknown settings: #{unknown_settings.join(', ')}" unless unknown_settings.empty?
  self._settings = self._settings.deep_merge(params)
end
```

pyromaniac: Why not? Also, http://apidock.com/rails/Hash/assert_valid_keys

pyromaniac:
I think `settings` name is too common, `default_import_options` maybe?

barthez:
@pyromaniac Is it better now?

pyromaniac:
Yeah, much better, thanks!

PR #354, opened 2016-03-23
Fixed invalid multibyte char error

Bajena:
This log lines caused SyntaxError: /usr/local/rvm/gems/jruby-1.7.22/gems/chewy-0.8.4/lib/chewy/type/witchcraft.rb:25: invalid multibyte char (US-ASCII)


pyromaniac:
:(

Thanks anyway!

PR #360, opened 2016-03-31
Reset import opts

cokuyan:
Allows the `#reset!` method to accept an options hash, specifically to allow the `#import` method to implement those options. This fork was originally created to allow passing in the `batch_size` option to the inner `#import` method.

### DIFF
--- a/lib/chewy/index/actions.rb
+++ b/lib/chewy/index/actions.rb
@@ -161,7 +161,7 @@ module Chewy
         #   UsersIndex.reset! Time.now.to_i
         #
         def reset! suffix = nil, opts = {}
-          import_opts = {suffix: suffix}.merge(opts[:import_opts])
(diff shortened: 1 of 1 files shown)


pyromaniac:
Well, right now you are able to specify import options per type, don't see why would you need different options for usual importing and reset

cokuyan:
What do you mean?
We needed to find a way to use the `reset!` method since it allows for creating an alias while the indexing occurs, whereas just doing a normal `import` would not create an alias during indexing.

pyromaniac:
Reset uses #import, #import uses https://github.com/toptal/chewy#type-default-import-options

cokuyan:
O, I didn't see that before. That definitely should solve the issue we had. Thanks!

pyromaniac:
So, did it help, can we close this PR?

cokuyan:
yea, go ahead

PR #363, opened 2016-04-01
Add method Index.type(type_name) which raises UndefinedType on unknown type_name

olleolleolle:
**Current behavior:**

undefined method `new' for nil:NilClass /usr/local/lib/ruby/gems/2.3.0/bundler/gems/chewy-27a71cc1ed55/lib/chewy/query.rb:1003:in `block in _results'

**New behavior:**

Raises `UndefinedType` with what Index was missing which Type.

```ruby
UsersIndex.type('user') # => UsersIndex::User
UsersIndex.type('foo') # => raises error UndefinedType("Unknown type in UsersIndex: foo")
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,6 +416,8 @@ You can access index-defined types with the following API:
 ```ruby
 UsersIndex::User # => UsersIndex::User
 UsersIndex.type_hash['user'] # => UsersIndex::User
+UsersIndex.type('user') # => UsersIndex::User
+UsersIndex.type('foo') # => raises error UndefinedType("Unknown type in UsersIndex: foo")
 UsersIndex.types # => [UsersIndex::User]
 UsersIndex.type_names # => ['user']
 ```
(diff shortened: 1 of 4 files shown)


pyromaniac:
Maybe it make sense to add a spec?

olleolleolle:
While adding a spec, I found I wanted a new method.

I found that I'd like to move this Hash-fetching-by-typename into a new method on the Index. The Index owns the `type_hash`. It should be responsible for querying it, and failing expressively when not finding things. 

Name proposal: `#lookup_type(name)`

pyromaniac:
Well, why not. Let's do it. I would simply call it `#type(name)`, wdyt?

olleolleolle:
@pyromaniac How about now?

pyromaniac:
Much better, going to merge this as soon as specs pass

pyromaniac:
ok, are we ready?

pyromaniac:
ah, ci didn't finish

olleolleolle:
Please do merge it all into a single commit - nothing in there worth keeping for my sake.

PR #372, opened 2016-04-15
Advanced Crutches™

pyromaniac:


lib/chewy/type/crutch.rb
```diff
@@ -23,7 +24,11 @@ def #{name}
 
       module ClassMethods
         def crutch name, &block
-          self._crutches = _crutches.merge(name.to_sym => block)
+          if block
+            self._crutches = _crutches.merge(name.to_sym => block)
+          else
+            self._crutches_classes = _crutches_classes + [name]
```
sergey-kintsel: why not just `<< name`?

pyromaniac: This is a class_attribute, we have to assign different object every time otherwise, it will be rewritten by ancestors

barthez: Why not:
```
self._crutches = _crutches.merge(name_or_class.name.to_sym => -> (collection) {
  name_or_class.new(collection)
})
```

pyromaniac: because then there will be a lot of instances, but we need only one

pyromaniac: btw, this is a great point, do we really need this change? I mean, we are able to do it in exactly this way without any patches

barthez: Aren't crutches memoized? We will have one instance per batch. his change would allow simplify crutches definition I guess.

additionaly If we could support something like:
```
crutch CrutchClass
...
field :name, crutch: [:crutch_class, :crutch_method] # translated to value: -> (item, crutch) {crutch.crutch_class.crutch_method(item) }
```

pyromaniac: Nope, this is too complicated with names, don't think it worth it

barthez: So I don't see any point of whole change :wink: 
As far as i can see it will only allow us to change this:
`value: -> (x, crutch) { crutch.something[x.id] }`
into: `value: -> (x, _, something) { something.else(x) }`

pyromaniac: Yeah, agree, let's go without this change. It is pointless.

barthez: @DNNX Had a better idea than that I suggested (with `crutch: 'name#name'`). Let's wait for his input here.

README.md
```diff
@@ -360,6 +360,34 @@ end
...
+
+There is one more option of crutches usage - you are able to define a crutch class:
+
+```ruby
+class ProductCrutch
+  def initialize(type, collection)
+    @collection = collection.map(&:id)
+  end
+
+  def categoties(product_id)
```
cedricpim: `categories`

README.md
```diff
@@ -360,6 +360,34 @@ end
...
+    crutch ProductCrutch
+
+    field :name
+    # simply use crutch-fetched data as a value:
+    field :category_names, value: ->(product, _crutches, product_crutch) { product_crutch.categories(product.id) }
+  end
+end
+```
+
+So the second crutch in the block will be an instance of `ProductCrutch`. You are able to add as many crutch classes as you need. The only thing to keep in mind here: all the block crutch definitions will go to the first crutch argument in the value proc, blockless classes provided will be instantiated additionaly and passed as following arguments. So you are able to mix block crutches and custom classes.
```
cedricpim: additionally

lib/chewy/type/crutch.rb
DNNX: Yea, I had the comment somewhere. I like this design more that the one which is proposed in the current PR:

https://github.com/toptal/platform/pull/9800#issuecomment-188677166

It is completely backwards compatible with the current interface of crutches, it's footprint is very small (the diff is gonna be short), it can be extended to using named crutches, and there are a bunch of other advantages described in the comment I mentioned.

> The only thing to keep in mind here: all the block crutch definitions will go to the first crutch argument in the value proc, blockless classes provided will be instantiated additionaly and passed as following arguments.

This ^ is kinda too hard to understand.

pyromaniac: I actually have much better interface in my mind and we definitely have to discuss it.

PR #375, opened 2016-04-21
Add support for raw_import flag

DNNX:
A new `raw_import` option was added. Basically, it allows to avoid unnecessary serialization->deserialization cycles when it matters. Often, a lot of time is spent on creating AR objects, typecasting fields, etc. But eventually, it all gets converted back to strings. So why waste resources on unnecessary conversions. Raw Import is added just for this purpose. On one of our indexes, using this feature shanked index reset time from 7.5 minutes to 4.5 minutes.

The details on how to use this feature can be found [in the updated documentation](https://github.com/toptal/chewy/pull/375/files#diff-04c6e90faac2675aa89e2176d2eec7d8R412).

### DIFF
--- a/lib/chewy/type/adapter/active_record.rb
+++ b/lib/chewy/type/adapter/active_record.rb
@@ -29,7 +29,13 @@ module Chewy
           result = true
 
           while ids.present?
-            result &= yield grouped_objects(default_scope_where_ids_in(ids))
+            objects =
+              if options[:raw_import]
+                raw_default_scope_where_ids_in(ids, options[:raw_import])
+              else
+                default_scope_where_ids_in(ids)
+              end
+            result &= yield grouped_objects(objects)
             break if ids.size < options[:batch_size]
             ids = pluck_ids(scope.where(target_id.gt(ids.last)))
           end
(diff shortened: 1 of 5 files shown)


lib/chewy/type/adapter/active_record.rb
```diff
@@ -49,6 +55,17 @@ def scope_where_ids_in(scope, ids)
           scope.where(target_id.in(Array.wrap(ids)))
         end
 
+        def raw_default_scope_where_ids_in(ids)
+          sql = default_scope.where(target_id.in(Array.wrap(ids))).to_sql
```
pyromaniac: `default_scope_where_ids_in(ids).to_sql`?

DNNX:
@pyromaniac please have a look

perlun:
@DNNX - thanks for a great feature. I'm probably a bit stupid (very newbie on ActiveRecord et al) but I am struggling to get it working based on the information you wrote there in the docs, but I cannot really understand how to get my head around it.

I have my ActiveRecord model set up etc, and via the [activerecord-import](https://github.com/zdennis/activerecord-import) gem an extra `#import` method is provided (which allows you to do do bulk-import into e.g. Postgres databases, which in turn speeds things up quite significantly). But how does the Chewy import method fit into the whole picture? Any suggestions would be highly appreciated.

PR #380, opened 2016-04-26
Add a weight scoring function to the search

sevab:
Example usage from the docs:
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html


pyromaniac:
Great, thanks!

PR #382, opened 2016-05-02
Refactor/remove memoization of index name prefix

JanBussieck:
There is an open pull request #314 concerning the issue that for multitenant applications the index prefix needs to be set dynamically (#225) which is currently not possible since the index name is memoized. I think there has been some misunderstanding concerning the point of the memoization which is not performance related, but rather a result of the fact that `index_name` is both a getter and setter (unfortunately) and it should be memoized if set explicitly.

This behavior is preserved in the changes I propose by saving the `index_name_stem` when it is set while allowing the prefix to be reevaluated every time `index_name` is accessed.

### DIFF
--- a/lib/chewy/index.rb
+++ b/lib/chewy/index.rb
@@ -29,11 +29,8 @@ module Chewy
     #   UsersIndex.index_name # => 'dudes'
     #
     def self.index_name(suggest = nil)
-      if suggest
-        @index_name_stem = suggest
-      end
-      @index_name = build_index_name(index_name_stem, prefix: default_prefix)
-      @index_name or raise UndefinedIndex
+      index_name = build_index_name(index_name_stem(suggest), prefix: default_prefix)
+      index_name or raise UndefinedIndex
     end
 
     # Prefix to use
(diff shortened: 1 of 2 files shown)


pyromaniac:
This code looks really cryptic, the control flow is spread all over the source. I would prefer to have less side-effects in methods.

JanBussieck:
You are right of course, I moved the memoization into the private method `index_name_stem`. We'll have the awkward behavior of either setting or computing a value in a single method either way, I moved this down to `index_name_stem` which is the actual value we want to memoize.

pyromaniac:
Alright, it looks much better. Could I ask you to add specs to consolidate this behavior? Also I'm going to think about this a little bit more over the weekend.

JanBussieck:
Sure will do. If you have any other suggestions concerning this change, I'd be happy to incorporate them.

JanBussieck:
Any idea why this little change causes so much breakage in other places?

pyromaniac:
Because everything else is counting on the name cache, I'm still going to think how to make it the best way, unfortunately I didn't have a chance yet :(

pyromaniac:
In master already

PR #384, opened 2016-05-06
Docs: Create a Table of Content

jesjos:
I did this one by hand. Would make sense to automate. But presented for you consideration like this.


pyromaniac:
You are awesome! Thanks!

Issue #383, opened 2016-05-03
how can i explicitly define the key field for model or exclude it from indexing at all?


okliv:
in addition to classes described here https://github.com/jeremyevans/sequel/issues/1178

i have this index model

    class ProductsIndex < Chewy::Index
    
      Product.subclasses.each do |subclass|
        define_type subclass do
          subclass.columns.each do |col|
              field col
          end
        end
      end
    end

and so if i try to 

    >> ProductsIndex.reset!

i receive

    Sequel::DatabaseError: Mysql2::Error: Column 'id' in field list is ambiguous

as far as i can understand, it is not strictly the sequel gem case.

i can imagine it can be fixed with somehow explicit `primary_key` definition to retrieve the row from DB because now `chewy` uses `id` attribute implicitly

pyromaniac:
Well, this requires a patch to sequel adapter, this code is deep inside and not configurable unfortunately.

okliv:
patch - is better than nothing if you could direct me in correct direction, i will try it... can you point file(s)/row(s) where i need to figure things out?

pyromaniac:
You may try to redefine this method correctly: https://github.com/toptal/chewy/blob/master/lib/chewy/type/adapter/sequel.rb#L42 to return id with table name instead of simply id. Also, it makes sense to understand which query exactly produces the exception: https://github.com/toptal/chewy/blob/master/lib/chewy/type/adapter/sequel.rb#L54 or https://github.com/toptal/chewy/blob/master/lib/chewy/type/adapter/sequel.rb#L50.

okliv:
yeah! i already digged this rows

    module Chewy
      class Type
        module Adapter
          class Sequel < Orm

            private

            def primary_key
              "#{target.table_name}__#{target.primary_key}"
            end
          end
      
        end
      end
   end

in my particular case it helped! now models are indexing...

but is it really ugly... any advices for better solution? i would try to develop/implement/contribute something more "stylish" tomorrow

pyromaniac:
Are you sure your patch doesn't break https://github.com/toptal/chewy/blob/master/lib/chewy/type/adapter/sequel.rb#L35 ?

I would say this solution is nice enough, just make sure all the specs are passing.

okliv:
i have requested a pull, but during tests one weird (for me) error appeared, so i do not know what to do with that (*elasticsearch didn't start because of some "permissions denied"*)

i can close this issue because now for me everything works

PR #385, opened 2016-05-06
complement primary_key to consider table name for correct behaviour for class table inherited requests

okliv:
#383


pyromaniac:
Awesome, thanks!

PR #387, opened 2016-05-18
attribute_highlight should return an array of highlights

musaffa:
Highlight of an attribute, for example `title_attribute`, should return an array rather than the first element of the array.

### DIFF
--- a/lib/chewy/type/wrapper.rb
+++ b/lib/chewy/type/wrapper.rb
@@ -42,7 +42,7 @@ module Chewy
       end
 
       def highlight(attribute)
-        _data["highlight"][attribute].first
+        _data["highlight"][attribute]
       end
 
       def highlight?(attribute)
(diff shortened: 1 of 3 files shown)


pyromaniac:
Hey, this is actually a good idea. But what about backward compatibility? Do you have any ideas? I meant, this feature is probably used by somebody already as is.

musaffa:
It's hard to add any deprecation warning. Perhaps it can go to 0.9 as minor version release or 1.0 as major release. The latter choice will be more semantic version compliant.

pyromaniac:
Right, So I'll postpone merging of this a little bit, ok?

musaffa:
Sure. How about adding this to a milestone?

pyromaniac:
You know what? I've though that we can call it `highlights`, `name_highlights`. This will allow not to break the existing API. WDYT?

musaffa:
`highlights` will be more appropriate I think.

pyromaniac:
We will probably need both

dalthon:
@musaffa thanks for your PR! It was rebased and merged at https://github.com/toptal/chewy/pull/755
We used `highlights` as suggested for new functionality and kept old behaviour for `highlight`

Issue #356, opened 2016-03-25
Regression: order(nil) no longer works


levent:
Regression in 0.8.4

In 0.8.3 order(nil) would silently work. Now it tries to pass that onto ES resulting in a valid exception.
I'm not sure if this is/was expected behaviour but thought I'd raise it just in case. I think it's common for users to do something like...

```ruby
UsersIndex.query(term: {name: 'foo'}).order(params[:order]) # sanitized params ideally where it could be left blank and result in nil.
```

```bash
[400] {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"malformed sort format, within the sort array, an object, or an actual string are allowed"}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":"test_app","node":"_apI_g7gQ4WT-vbWdiqyvQ","reason":{"type":"illegal_argument_exception","reason":"malformed sort format, within the sort array, an object, or an actual string are allowed"}}]},"status":400}
```

pyromaniac:
I don't remember I did something with DSL between releases, did you check elasticsearch-ruby for changed behavior?

levent:
No other gems or ES itself have changed.
I'll try and investigate further.

pyromaniac:
Would be perfect, thanks!

levent:
This is a very crude experiment but I forked both tags (v0.8.4 and v0.8.3) and changed one line in the kaminari spec: https://github.com/toptal/chewy/commit/0aebb6f3a2970f5922cdb2f6f6ab4a046006aeeb

Tests in v0.8.3 - https://travis-ci.org/levent/chewy/jobs/118462121
Tests in v0.8.4 - https://travis-ci.org/levent/chewy/jobs/118461886

pyromaniac:
Could I ask you to submit a fix please? I frankly have no idea when will I be able to fix this, definitely not earlier the next weekend. Yeah, regression seems to be exists somehow :( Have no idea how though.

levent:
I'll try and find some time this week :+1:

pyromaniac:
Would be great, thanks!

levent:
Apologies for the delay!

I finally found the issue. I wonder what you propose?

See https://github.com/toptal/chewy/blob/master/lib/chewy/query/criteria.rb#L117
```sort?``` returns ```true``` when the value of ```sort``` is ```[nil]```

[...]

PR #388, opened 2016-05-20
Fix regression where order(nil) would be passed through as invalid sort

levent:
https://github.com/toptal/chewy/issues/356


pyromaniac:
Awesome, thanks!

Issue #338, opened 2016-02-25
Multi clients support


mainameiz:
I want to implement multi-client support as it is in mongoid where you can specify different hosts for different collections.

E.g.

``` yaml
# chewy.yml

production:
  clients:
    default:
      hosts: "localhost:9200"
    search:
      hosts: "localhost:9201"
```

``` ruby
# app/chewy/search_index.rb

class SearchIndex < Chewy::Index
  client :search
end
```

Is it a good idea? Do you have any suggestions about this feature?

pyromaniac:
Well, why not, suppose it will be useful. But keep configuration backward-compatibility please

mainameiz:
Done https://github.com/toptal/chewy/pull/391

mainameiz:
Sorry for taking so much of your time @pyromaniac. We are going to split our application into two services and don't need this feature anymore. I am going to close this issue and all related PRs as they are stale.

PR #391, opened 2016-05-22
Support for multiple hosts.

mainameiz:
Adds an ability to use different hosts for different indexes. E.g.

Example of config:

``` ruby
production:
   hosts: 'localhost:9250' # default host for all indexes
   clients: # alternative host
     search:
       hosts: 'search.example.com:9250'
```

Specify it in index.

``` ruby
class SearchIndex < Chewy::Index
   use_client :search
end
```


AndreySavelyev:
👍

pyromaniac:
Sorry for not paying attention to this awesome PR, but I'm quite busy currently, I'll try to find a time for it

lib/chewy/index.rb
```diff
@@ -30,22 +35,31 @@ class Index
...
 
-    # Prefix to use
+    def self.use_client(name)
+      #puts "use_client: #{name}"
+      self.client_name = name.to_sym
+    end
+
+    # Elasticsearch::Client to use
     #
-    def self.default_prefix
```
pyromaniac: Why did you get rid of this method? It was explicitly added in order to be able to change the prefix from every index. It was like a feature :)

lib/chewy/clients.rb
```diff
@@ -0,0 +1,37 @@
...
+    def self.clear
+      clients.clear
+    end
+
+    def self.default
+      with_name(:default)
+    end
+
+    def self.with_name(name)
+      #puts "Chewy::Clients.with_name: #{name}"
```
mainameiz: remove

lib/chewy/config.rb
```diff
@@ -94,6 +94,20 @@ def transport_tracer= tracer
...
+    #      :clients - configuration for alternative elastic search servers.
+    #
+    #        test:
+    #          hosts: 'localhost:9250'
+    #          clients:
+    #            search:
+    #              hosts: 'search.example.com:9250'
+    #
+    #         class SearchIndex < Chewy::Index
+    #           client :search
```
mainameiz: `use_client :search`

mainameiz:
Ok, whatever it seems to me this PR can and should be improved)

mykolas921:
Hi @pyromaniac, @mainameiz,

I need this PR badly. When do you this PR can be merged?

Thanks,
Mykolas

pyromaniac:
Hey, well, as soon as it will be ready. You may try to propose your updates here

mainameiz:
@pyromaniac Hello! I have rebased it with master branch and fix errors. Could you review it?

pyromaniac:
Hey man. This is a great improvement. I reviewed it and it definitely deserves to be merged. The only thing here is that I would like to refactor some code there. For example, it makes sense to create configurations as instances of a class, something like:

```ruby
class Config
  def client; ... end

  def settings; ... end
end

Chewy.config(:default).client
Chewy.config(:server2).settings
Chewy.config(:server2).settings = { additional: 'stuff' }
```

I want to refactor this piece of code for a long time already, but I want to create a great interface here. If you will be unable to do it - I'll join you as soon as will finish with the ES 5 support I'm working on currently. WDYT?

mainameiz:
@pyromaniac I can do this refactoring. So, as I understand, you want:
- add `Chewy::Config#client` method instead of `Chewy::Clients` class
- add ability to pass `config_name` parameter to `Chewy.config(config_name)` instead of `Chewy::Configs.with_name(config_name)`

pyromaniac:
Yeah, let's start with this and see how would it look like.

pyromaniac:
Hm, we've got Config already, probably it makes sense to split it, keep global options there and mode the settings thingy to a new class Chewy::Environment or maybe think about a better name.

mainameiz:
@pyromaniac done. Please, find a minute to review it.

It becomes much larger, but I have found more places where a code must be a client-specific and fixed it.

And I would like to ask you a question - how about to broke some things and place default config under `clients:` config key as it done in `mongoid`?

```yaml
development:
  # Configure available database clients. (required)
  clients:
    # Defines the default client. (required)
    default:
      # Defines the name of the default database that Mongoid can connect to.
      # (required).
      database: '***'
```

mainameiz:
I will resolve conflicts a little bit later

pyromaniac:
Didn't get, what's going to be on the top level then?

mainameiz:
@pyromaniac

Changes:

- `Chewy.client` now requires name of client (previous version is deprecated). Usages of `Chewy.client` replaced with `Chewy.default_client` and `Chewy.client(client_name)`. It uses `Chewy::Clients` underneath.
- `Chewy.clients[client_name]` returns config for specified client instead of `Chewy::Configs.with_name(client_name)` but it still uses `Chewy::Config::Clients` underneath.
- `Chewy::Runtime.version` replaced with `Chewy.client(client_name).version`
- `Chewy.default_client` returns instance of `Chewy::Client` instead of `Elasticsearch::Client` (it is wrapper around `Elasticsearch::Client` with delegation of methods).

> Hm, we've got Config already, probably it makes sense to split it, keep global options there and move the settings thingy to a new class Chewy::Environment or maybe think about a better name.

As it seems to me, `Config` is ok. Why you think it should be splitted?

mainameiz:
@pyromaniac 

> Didn't get, what's going to be on the top level then?

Top level will contains default parameters for `query_mode`, `filter_mode` etc.

pyromaniac:
Ok, if you will take a look at the current config class, you will see that it contains the settings hash and some global configuration options such as `indices_path` or `request_strategy`.

Settings also consist of:
1. Client options - which totally should be in different YAML branches
2. Prefix - this one should be configured per client, but it actually makes sense to have it configurable globally.
3. Index default settings (are merged to every index settings hash, used on index creation) - this one should behave the same way as the prefix, I believe.

While `indices_path`, for example, should be configured globally only, not per-client, the same thing is with the `request_strategy` or DSL settings. As for them, btw, I'm introducing new DSL, so it makes sense to get rid of `query_mode` from the config and put it into the query-class specific namespace, and I'll probably do this in my branch.

That's why I want to split Config. Because it was `Chewy.config.indices_path` and `Chewy.config.settings`, but should become: `Chewy.config.indices_path` (stay the same) and `Chewy.client(:client_name).settings`. It also makes sense to think, about better naming maybe.

So I was imagining this interface the following way:

* `Chewy.client(:client_name).client_settings`
* `Chewy.config(:client_name).index_settings`
* `Chewy.config(:client_name).prefix`
* `Chewy.config(:client_name).version`
* `Chewy.config(:client_name).client`

Basically, here, first 3 options are replacing the current `Chewy.config.settings`. But maybe it doesn't worth to split them and makes sense to keep: `Chewy.config(:client_name).settings`

The thing I don't like right now in my idea is that there is no clean interface for `indices_path`, for example, we will have to use `Chewy.config.indices_path`, which is confusing. And I like your idea about the client wrapper, however not sure if it is not an overcomplication. So better interface in my opinion will look like:

* `Chewy.client(:client_name).client_settings`
* `Chewy.client(:client_name).index_settings`
* `Chewy.client(:client_name).prefix`
* `Chewy.client(:client_name).version`
* `Chewy.client(:client_name).whatever_elasticsearch_client_method_proxified`

In this case, global config will stay in the `.config` space: `Chewy.config.indices_path`, but still, you see, `Chewy::Config` class will be cleaned up from `settings` or prefixes.

Also maybe we can go without proxy and use `Chewy.client(:client_name).real_client` or `Chewy.client(:client_name).es_client`, or maybe we will fing a better global accessor and will be able to use `Chewy.better_name(:client_name).client`.

Anyway, I want `Chewy.client(:client_name).client_settings` to be settable also from the ruby code, not only from YAML. The same story for `Chewy.client(:client_name).index_settings` or `Chewy.client(:client_name).prefix`.

Also, I don't want `Chewy.config.indices_path` or other global things to be settable from YAML, it doesn't make sense as for me.

So I think the YAML structure have to provide a way to set up prefixes or some default stuff like default index settings or even client setting (port, for example, however not sure, does it worth it or not) globally. In this case, your proposal makes total sense:

```yaml
development:
  prefix: 'global_prefix'
  # Configure available database clients. (required)
  clients:
    # Defines the default client. (required)
    default:
      # Defines the name of the default database that Mongoid can connect to.
      # (required).
      prefix: 'prefix_for_default_client'
```

Sorry for such a long explanation I'm just trying to provide my understanding to come to the best solution together.

mainameiz:
@pyromaniac So you give me a permission to do these changes? I'm trying not to break APIs as you say (https://github.com/toptal/chewy/issues/338).
It is better to do in this PR or another? It seems to be quite later already...

pyromaniac:
It is a good point about the backward compatibility. In #338 I was actually talking about the YAML reader backward compatibility. I have to think about it. I want to change the API completely just in order to make it nice and clear and if we are going to keep the API backward compatibility - let's make it separate completely like we will have 2 completely different APIs without having methods with the double purpose. Let's come up with good naming and structure for it, I'm going to think a little bit.

stephencattaneo:
I'm very interested in this work. The last update for the issue was a month and a half ago.  An expectation that the PR will merged into master anytime soon?

mainameiz:
@pyromaniac Hello! I want to continue to work on this feature. It's is time to update my elasticsearch cluster to v5 :) So I need `chewy`'s changes from master branch.

What should I do to get this PR merged?

pyromaniac:
First of all, let's figure out the backward-compatible interface. And then just implement it. Can you produce the list of requirements you think are necessary? So I will do as well.

mainameiz:
First of all let's rename `Chewy.client` to `Chewy.default_client` everywhere in codebase (I will do separate PR for this) and deprecate `Chewy.client` method. So it will reduce changes in next PRs.

mainameiz:
You sad:

> That's why I want to split Config

But we can't split config without breaking compability. So here I need you decision:
1) break compability and split configs
2) keep compability and add just add `:clients` key to yml config in which other configs will be defined as in top level.

```
production:
  ... top level parameters for `Chewy.default_client` (hosts, wait_for_status, etc.)
  clients:
    es2_client:
      ... same parameters as at top level (hosts, wait_for_status, etc.) but without inheritance
```

pyromaniac:
I would use the second one, but deprecate the top-level config in favor of

```
clients:
  default:
```

The same I would do with the Ruby API. Just keep the current one but deprecate it and keep as an alias to the brand new API we will develop.

mainameiz:
Like this https://github.com/toptal/chewy/pull/613?

jbmeerkat:
Up! Are there any chances to merge this?

mainameiz:
https://github.com/toptal/chewy/issues/338

PR #393, opened 2016-05-27
Add unlimited method to fetch all records from ES

sergey-kintsel:
It removes offset and sets response limit to the total_count value

### DIFF
--- a/lib/chewy/query.rb
+++ b/lib/chewy/query.rb
@@ -289,8 +289,8 @@ module Chewy
     #            size: 100
     #          }}
     #
-    def limit value
-      chain { criteria.update_request_options size: Integer(value) }
+    def limit value = nil, &block
+      chain { criteria.update_request_options size: block || Integer(value) }
     end
 
     # Sets elasticsearch <tt>from</tt> search request param
@@ -301,8 +301,8 @@ module Chewy
     #            from: 300
     #          }}
     #
-    def offset value
-      chain { criteria.update_request_options from: Integer(value) }
+    def offset value = nil, &block
+      chain { criteria.update_request_options from: block || Integer(value) }
     end
 
     # Elasticsearch highlight query option support
@@ -894,8 +894,8 @@ module Chewy
     #   scope = UsersIndex.aggs(max_age: { max: { field: 'age' } }).search_type(:count)
     #   max_age = scope.aggs['max_age']['value']
     #
-    def search_type val
-      chain { options.merge!(search_type: val) }
+    def search_type value
+      chain { criteria.update_search_options search_type: value }
     end
 
     # Merges two queries.
@@ -961,6 +961,16 @@ module Chewy
       search_type(:count).total > 0
     end
 
+    # Sets limit to be equal to total documents count
+    #
+    #  PlacesIndex.query(...).filter(...).unlimited
+    #
+
+    def unlimited
+      count_query = search_type(:count)
+      offset(0).limit { count_query.total }
+    end
+
     # Returns request total time elapsed as reported by elasticsearch
     #
     #   UsersIndex.query(...).filter(...).took
@@ -1004,7 +1014,6 @@ module Chewy
       @_request ||= begin
         request = criteria.request_body
         request.merge!(index: _indexes.map(&:index_name), type: _types.map(&:type_name))
-        request.merge!(search_type: options[:search_type]) if options[:search_type]
         request
       end
     end
--- a/lib/chewy/query/criteria.rb
+++ b/lib/chewy/query/criteria.rb
@@ -5,7 +5,7 @@ module Chewy
     class Criteria
       include Compose
       ARRAY_STORAGES = [:queries, :filters, :post_filters, :sort, :fields, :types, :scores]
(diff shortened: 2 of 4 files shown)


lib/chewy/query/criteria.rb
```diff
@@ -5,7 +5,7 @@ class Query
     class Criteria
       include Compose
       ARRAY_STORAGES = [:queries, :filters, :post_filters, :sort, :fields, :types, :scores]
-      HASH_STORAGES = [:options, :request_options, :facets, :aggregations, :suggest, :script_fields]
+      HASH_STORAGES = [:options, :request_options, :facets, :aggregations, :suggest, :script_fields, :search_options]
```
pyromaniac: move it before request_options please

lib/chewy/query.rb
```diff
@@ -1004,7 +1014,7 @@ def _request
       @_request ||= begin
         request = criteria.request_body
         request.merge!(index: _indexes.map(&:index_name), type: _types.map(&:type_name))
-        request.merge!(search_type: options[:search_type]) if options[:search_type]
+        request.merge!(search_type: criteria.search_options[:search_type]) if criteria.search_options[:search_type]
```
pyromaniac: move this logic to `request_body` method, like `{body: ...}.merge(serch_options)`

lib/chewy/query.rb
```diff
@@ -961,6 +961,16 @@ def exists?
...
     end
 
+    # Sets limit to be equal to total documents count
+    #
+    #  PlacesIndex.query(...).filter(...).unlimited
+    #
+
+    def unlimited
+      size = search_type(:count).total
+      chain { criteria.update_request_options size: Integer(size), from: 0 }
```
pyromaniac: Is size a string here?

spec/chewy/query_spec.rb
```diff
@@ -96,13 +96,15 @@
     specify { expect(subject.limit(10)).not_to eq(subject) }
     specify { expect(subject.limit(10).criteria.request_options).to include(size: 10) }
     specify { expect { subject.limit(10) }.not_to change { subject.criteria.request_options } }
+    specify { expect(subject.offset { 20/2 }.criteria.request_body[:body]).to include(size: 10) }
```
pyromaniac: This shell not pass!

spec/chewy/query_spec.rb
```diff
@@ -454,7 +456,7 @@
   end
 
   describe '#exists?' do
-    let(:data) { 10.times.map { |i| {id: i.next.to_s, name: "Name#{i.next}", age: 10 * i.next}.stringify_keys! } }
+    let(:data) { Array.new(10) { |i| {id: i.next.to_s, name: "Name#{i.next}", age: 10 * i.next}.stringify_keys! } }
```
pyromaniac: what's bad about `10.times.map`? :(

bbatsov: It's slower. :-)

pyromaniac: Oh, really? Don't think it is significant in this particular case, `10.times.map` looks more like ruby-way as for me, `Array.new(10)` is c++ or java.

bbatsov: True.

On Friday, 27 May 2016, Arkadiy Zabazhanov <notifications@github.com> wrote:

> In spec/chewy/query_spec.rb
> <https://github.com/toptal/chewy/pull/393#discussion_r64883611>:
>
> > @@ -454,7 +456,7 @@
> >    end
> >
> >    describe '#exists?' do
> > -    let(:data) { 10.times.map { |i| {id: i.next.to_s, name: "Name#{i.next}", age: 10 * i.next}.stringify_keys! } }
> > +    let(:data) { Array.new(10) { |i| {id: i.next.to_s, name: "Name#{i.next}", age: 10 * i.next}.stringify_keys! } }
>
> Oh, really? Don't think it is significant in this particular case,
> 10.times.map looks more like ruby-way as for me, Array.new(10) is c++ or
> java.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/toptal/chewy/pull/393/files/8a9a6bbeb01054ef878640b595be7e7da2298f8c#r64883611>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAGVynrZ3G_eFfzWT94Z0GGjfaG-HtWHks5qFsEDgaJpZM4IoK68>
> .
>


-- 
Cheers,
Bozhidar

Bozhidar Batsov, VP of Engineering || Toptal <https://toptal.com> ||
bozhidar_ (skype)

pyromaniac:
Awesome! Thanks!

binarydev:
@sergey-kintsel does this require ES 5.x or a special max_results_window value? Tried to use it locally with my ES 2.3 installation, and it complains about going beyond the max results window

sergey-kintsel:
@binarydev Hi! It does not require the latest ES, but it does require max_results_window. The problem is that most likely you are trying to load too much data. This method is a bit unsafe I would say. So, if you really need that much data - change max results window

PR #398, opened 2016-06-09
Fix routing_missing_exception on delete with parent missing

guigs:
Elasticsearch 2.2+ returns error when trying to delete an object that must have parent but parent is not sent.

This occurs when object is not present, so existing_object is nil. In this case, it was not populating parent key and then causing error routing_missing_exception on Elasticsearch.

The solution is to skip delete if record does not exists.

### DIFF
--- a/lib/chewy/type/import.rb
+++ b/lib/chewy/type/import.rb
@@ -121,7 +121,8 @@ module Chewy
 
           if root_object.parent_id
             existing_object = entry[:_id].present? && indexed_objects && indexed_objects[entry[:_id].to_s]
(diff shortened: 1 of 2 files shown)


guigs:
I'll try to add a test case later.

pyromaniac:
Hey! Is this ready to be merged? I see a strange duplication in specs.

guigs:
Yes, the duplication is on purpose. The second call would not pass before the fix as explained in commit message of acc8864e5b9b105d3d313e1c4f178e119db9f6ae.

pyromaniac:
Perfect, thanks!

PR #407, opened 2016-07-25
Journaling

sergey-kintsel:
It adds journaling for all actions that were made on records. It is optional and can be set using import options, `default_import_options` and Chewy configuration. 

### TODO
- [x] Add changes applying
- [x] Add rake task for changes applying

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 gemspec
 
-# gem 'activerecord'
+gem 'activerecord'
 # gem 'mongoid'
 # gem 'sequel'
 
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem 'activerecord'
+# gem 'activerecord'
 # gem 'mongoid'
 # gem 'sequel'
 
--- a/lib/chewy.rb
+++ b/lib/chewy.rb
@@ -31,6 +31,7 @@ require 'chewy/index'
 require 'chewy/type'
 require 'chewy/fields/base'
 require 'chewy/fields/root'
+require 'chewy/journal'
 require 'chewy/railtie' if defined?(::Rails)
 
 begin
@@ -144,9 +145,8 @@ module Chewy
     # Be careful, if current prefix is blank, this will destroy all the indexes.
     #
     def massacre
-      result = Chewy.client.indices.delete(index: [Chewy.configuration[:prefix], '*'].delete_if(&:blank?).join(?_))
-      Chewy.wait_for_status if result
-      result
+      Chewy.client.indices.delete(index: [Chewy.configuration[:prefix], '*'].delete_if(&:blank?).join(?_))
+      Chewy.wait_for_status
     end
     alias_method :delete_all, :massacre
 
--- a/lib/chewy.rb
+++ b/lib/chewy.rb
@@ -144,8 +144,9 @@ module Chewy
     # Be careful, if current prefix is blank, this will destroy all the indexes.
     #
     def massacre
-      Chewy.client.indices.delete(index: [Chewy.configuration[:prefix], '*'].delete_if(&:blank?).join(?_))
-      Chewy.wait_for_status
+      result = Chewy.client.indices.delete(index: [Chewy.configuration[:prefix], '*'].delete_if(&:blank?).join(?_))
+      Chewy.wait_for_status if result
+      result
     end
     alias_method :delete_all, :massacre
 
--- a/lib/chewy/index/actions.rb
+++ b/lib/chewy/index/actions.rb
@@ -158,12 +158,12 @@ module Chewy
         # zero-downtime index resetting (described here:
         # http://www.elasticsearch.org/blog/changing-mapping-with-zero-downtime/).
         #
-        #   UsersIndex.reset! Time.now.to_i
+        #   UsersIndex.reset! Time.now.to_i, journal: true
         #
-        def reset! suffix = nil
+        def reset! suffix = nil, journal: false
           if suffix.present? && (indexes = self.indexes).present?
             create! suffix, alias: false
-            result = import suffix: suffix
+            result = import suffix: suffix, journal: journal
             client.indices.update_aliases body: {actions: [
               *indexes.map do |index|
                 {remove: {index: index, alias: index_name}}
@@ -174,7 +174,7 @@ module Chewy
             result
           else
             purge! suffix
-            import
+            import journal: journal
           end
         end
       end
--- a/lib/chewy/journal.rb
+++ b/lib/chewy/journal.rb
@@ -80,6 +80,11 @@ module Chewy
         result
       end
 
+      def delete!
+        Chewy.client.delete_by_query index: index_name, type: type_name, body: {query: {match_all: {}}}
+        Chewy.wait_for_status
+      end
+
       def exists?
         Chewy.client.indices.exists? index: index_name
       end
--- a/lib/chewy/journal.rb
+++ b/lib/chewy/journal.rb
@@ -1,0 +1,99 @@
+module Chewy
+  class Journal
+    JOURNAL_MAPPING = {
+      journal: {
+        properties: {
+          index_name: {type: 'string', index: 'not_analyzed'},
+          type_name: {type: 'string', index: 'not_analyzed'},
+          action: {type: 'string', index: 'not_analyzed'},
+          object_ids: {type: 'string', index: 'not_analyzed'},
+          created_at: {type: 'date', format: 'basic_date_time'}
+        }
+      }
+    }.freeze
+
+    def initialize(index)
+      @records = []
+      @index = index
+    end
+
+    def add(action_objects)
+      @records +=
+        action_objects.map do |action, objects|
+          {
+            index_name: @index.index_name,
+            type_name: @index.type_name,
+            action: action,
+            object_ids: identify(objects),
+            created_at: Time.now.to_i
+          }
+        end
+    end
+
+    def bulk_body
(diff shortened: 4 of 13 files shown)


lib/chewy/index/actions.rb
```diff
@@ -160,10 +160,10 @@ def #{method} options = {}
         #
         #   UsersIndex.reset! Time.now.to_i
         #
-        def reset! suffix = nil
+        def reset! suffix = nil, import_options = {}
```
pyromaniac: I would use extract_options! here

lib/chewy/journal.rb
```diff
@@ -0,0 +1,97 @@
...
+          index_name: {type: 'string', index: 'not_analyzed'},
+          type_name: {type: 'string', index: 'not_analyzed'},
+          action: {type: 'string', index: 'not_analyzed'},
+          object_ids: {type: 'string', index: 'not_analyzed'},
+          created_at: {type: 'date', format: 'basic_date_time'}
+        }
+      }
+    }.freeze
+
+    def initialize
```
pyromaniac: I would pass current adapter to the instance here

lib/chewy/type/import.rb
```diff
@@ -31,10 +32,13 @@ def import *args
 
           ActiveSupport::Notifications.instrument 'import_objects.chewy', type: self do |payload|
             adapter.import(*args, import_options) do |action_objects|
+              journal = Chewy::Journal.new
```
pyromaniac: Yeah, Chewy::Journal.new(self).add(action_objects)

lib/chewy/type/import.rb
```diff
@@ -14,9 +14,10 @@ module ClassMethods
         #   UsersIndex::User.import [1, 2, 3]                # imports users with specified ids
         #   UsersIndex::User.import users                    # imports users collection
         #   UsersIndex::User.import suffix: Time.now.to_i    # imports data to index with specified suffix if such exists
-        #   UsersIndex::User.import refresh: false           # to disable index refreshing after import
```
DNNX: Why this line was removed? Is this option becoming deprecated?

spec/chewy/journal_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+
+      Timecop.freeze(import_time)
+      PlacesIndex.import
+
+      expect(Chewy.client.indices.exists?(index: 'chewy_journal')).to eq true
+
+      Timecop.freeze(update_time)
+      cities.first.update(id: 9000)
+
+      Timecop.travel(destroy_time)
```
DNNX: Why `travel` here and `freeze` a few lines above?

lib/chewy/type/import.rb
```diff
@@ -14,9 +14,10 @@ module ClassMethods
...
+        #   UsersIndex::User.import journal: true            # import will record all the actions into special journal index
         #   UsersIndex::User.import batch_size: 300          # import batch size
         #   UsersIndex::User.import bulk_size: 10.megabytes  # import ElasticSearch bulk size in bytes
+        #   UsersIndex::User.import refresh: false           # to disable index refreshing after import
```
sergey-kintsel: @DNNX it wasn't 😃

lib/chewy/journal.rb
```diff
@@ -0,0 +1,97 @@
...
+        result
+      end
+
+      def exists?
+        Chewy.client.indices.exists? index: index_name
+      end
+
+      def index_name
+        [
+          Chewy.configuration[:prefix],
```
pyromaniac: Indexed have a personal prefix configuration, use it instead

lib/chewy/rake_helper.rb
```diff
@@ -47,7 +47,7 @@ def normalize_indexes *indexes
       def reset_index *indexes
         normalize_indexes(indexes).each do |index|
           puts "Resetting #{index}"
-          index.reset! (Time.now.to_f * 1000).round
+          index.reset! (Time.now.to_f * 1000).round, journal: false
```
pyromaniac: We can actually do journal: false a default option for the reset! method, but is should be redefinable for sure.

lib/chewy/journal.rb
pyromaniac: Ah you will have to move all this methods to the instance

lib/chewy/journal.rb
```diff
@@ -0,0 +1,97 @@
...
+            range: {
+              created_at: {
+                gte: time.to_i
+              }
+            }
+          }
+        }
+        Chewy.client
+          .search(index: index_name, type: type_name, body: query,
+                  sort: 'created_at')['hits']['hits']
```
pyromaniac: Fix this terrible indentations please

lib/chewy/journal.rb
```diff
@@ -0,0 +1,97 @@
...
+          .map { |r| r['_source'] }
+      end
+
+      def create
+        return if exists?
+        result = Chewy.client.indices.create index: index_name,
+                                             body: {settings: Chewy.settings,
```
pyromaniac: Fix this terrible indentations please

lib/chewy.rb
```diff
@@ -144,8 +145,9 @@ def wait_for_status
     # Be careful, if current prefix is blank, this will destroy all the indexes.
     #
     def massacre
-      Chewy.client.indices.delete(index: [Chewy.configuration[:prefix], '*'].delete_if(&:blank?).join(?_))
-      Chewy.wait_for_status
+      result = Chewy.client.indices.delete(index: [Chewy.configuration[:prefix], '*'].delete_if(&:blank?).join(?_))
+      Chewy.wait_for_status if result
```
pyromaniac: Are you sure about this change?

lib/chewy/journal.rb
pyromaniac: Well, actually I think is doesn't worth it, let's use a global prefix, let's don't mix stuff

pyromaniac:
It probably makes sense to test that reset! method doesn't journal anything by default.

sergey-kintsel:
okay

lib/chewy/type/import.rb
```diff
@@ -218,6 +228,10 @@ def fetch_indexed_objects(objects)
 
           indexed_objects
         end
+
+        def journal?(import_options)
+          import_options.key?(:journal) ? import_options[:journal] : Chewy.configuration[:journal]
```
DNNX: `import_options.fetch(:journal) { Chewy.configuration[:journal] }`

lib/chewy/journal.rb
```diff
@@ -0,0 +1,99 @@
...
+            created_at: Time.now.to_i
+          }
+        end
+    end
+
+    def bulk_body
+      @records.map do |record|
+        {
+          create: {
+            _index: Journal.index_name,
```
pyromaniac: self.class

lib/chewy/rake_helper.rb
```diff
@@ -47,7 +47,11 @@ def normalize_indexes *indexes
       def reset_index *indexes
         normalize_indexes(indexes).each do |index|
           puts "Resetting #{index}"
-          index.reset! (Time.now.to_f * 1000).round
+          time = Time.now
+          index.reset! (time.to_f * 1000).round
+          if index._default_import_options.fetch(:journal) { Chewy.configuration[:journal] }
```
pyromaniac: Why don't do a method like `index.journal?`

lib/chewy/type/import.rb
```diff
@@ -31,10 +32,13 @@ def import *args
 
           ActiveSupport::Notifications.instrument 'import_objects.chewy', type: self do |payload|
             adapter.import(*args, import_options) do |action_objects|
+              journal = Chewy::Journal.new(self)
+              journal.add(action_objects) if import_options[:journal] || journal?
```
pyromaniac: Still if import_options[:journal] will be false, it will not redefine the global config.
`import_options.fetch(:journal) { journal? }`

lib/chewy/type/import.rb
```diff
@@ -66,6 +70,7 @@ def bulk options = {}
           suffix = options.delete(:suffix)
           bulk_size = options.delete(:bulk_size)
           body = options.delete(:body)
+          journal = options.delete(:journal)
```
pyromaniac: do you need this line?

pyromaniac:
I would add a rake task for journal cleaning, it will be required right away.

lib/chewy/type/import.rb
sergey-kintsel: I think so because we use `options` below we're merging it with `bulk_body` and `header` and we definitely shouldn't provide some "garbage" to `Chewy.client.bulk`

pyromaniac: Ok, I've just found that delete works as destructive fetch, so you can simply do `journal.add(action_objects) if import_options.delete(:journal) { journal? }` above

pyromaniac: Ah, got it now, ok, those are two different methods. Never mind.

lib/chewy/type/import.rb
```diff
@@ -88,6 +93,11 @@ def bulk options = {}
             [body]
           end
 
+          if journal.any_records?
+            Chewy::Journal.create
```
pyromaniac: I believe this journal creation have to be in the `import method`

pyromaniac: Or maybe no, whatever

Gemfile
```diff
@@ -17,4 +17,5 @@ group :test do
   gem 'guard'
   gem 'guard-rspec'
   gem 'listen', '~> 3.0.0'
+  gem 'timecop'
```
pyromaniac: It would be much better to move timecop to the gemspec instead as a development dependency

PR #411, opened 2016-08-02
Load all entries from journal, not only 10 as set by default

sergey-kintsel:


lib/chewy/journal.rb
```diff
@@ -61,16 +61,20 @@ def apply_changes_from(time)
...
-                gte: time.to_i
-              }
+          range: {
+            created_at: {
+              gte: time.to_i
             }
           }
         }
-        Chewy.client.search(index: index_name, type: type_name, body: query, sort: 'created_at')['hits']['hits'].map { |r| r['_source'] }
+        size = Chewy.client.count(index: index_name, type: type_name, body: {query: query})['count']
```
pyromaniac: why do we have query here?

pyromaniac: Can't you use Chewy::Query ?

sergey-kintsel: because count doesn't accept `filter`: `QueryParsingException[[some_prefix_chewy_journal] request does not support [filter]]`

pyromaniac: okay

pyromaniac: this is strange however, I've got a feeling that something is wrong there

sergey-kintsel: can you share your thoughts? or you mean that it is strange that `count` doesn't support `filter`? if so then I agree. Maybe it's because I am using Elasticsearch 1.7.5. Maybe they have fixed it later

lib/chewy/journal.rb
```diff
@@ -87,9 +97,13 @@ def delete!
       def clean_until(time)
         query = {
           query: {
-            range: {
-              created_at: {
-                lte: time.to_i
+            filtered: {
+              query: {
```
pyromaniac: filter ?

sergey-kintsel: strange, but it worked. Thanks :+1:

pyromaniac: sure it worked, but less efficient

sergey-kintsel: I meant that it worked when I changed `query` to `filter`, but forget it. It doesn't. It makes my Elastic yellow

PR #413, opened 2016-08-05
Add grouping for Chewy journal entries

sergey-kintsel:


lib/chewy/journal.rb
```diff
@@ -118,5 +120,21 @@ def query(time, comparator, query_type = :filter)
         }
       end
     end
+
+    Entry = Struct.new(:index_name, :type_name, :action, :object_ids, :created_at) do
```
pyromaniac: JournalEntry and maybe makes sense to extract if to the own file

sergey-kintsel: No™

pyromaniac: You will regret about this.

sergey-kintsel: Maybe, but who knows what will be 😄

pyromaniac: I know

lib/chewy/journal.rb
```diff
@@ -118,5 +120,31 @@ def query(time, comparator, query_type = :filter)
...
       end
     end
+
+    class Entry
+      ATTRIBUTES = %w[index_name type_name action object_ids created_at].freeze
+
+      attr_accessor *ATTRIBUTES
+
+      def initialize(attributes = {})
+        attributes.slice(*ATTRIBUTES).each do |attr, value|
```
pyromaniac: attributes.stringify_keys.slice

PR #414, opened 2016-08-08
Allow app/chewy location to be changed via yml config

robacarp:
Hello again!

In our rails application directory we have the following folders:

```
assets/
contexts/
controllers/
errors/
exporters/
helpers/
jobs/
mailers/
models/
policies/
serializers/
views/
```

While integrating the Chewy gem, I added the `app/chewy` folder to the application. There were a few complaints that 'chewy' didn't match the convention applied in the `app/` directory, and so I was able to change it to `app/indices` with a little creative monkey patching.

This patch allows the directory to be set with a new config option called `index_definition_path`.

I tried to find every place the path was hard coded and replace with `Chewy.configuration[:index_definition_path]`.

Additionally, the default setting is `app/chewy` so current deployments shouldn't need to change anything.

Thoughts?

Thanks!

### DIFF
--- a/lib/chewy/config.rb
+++ b/lib/chewy/config.rb
@@ -40,7 +40,7 @@ module Chewy
 
       # Where Chewy expects to find index definitions
       # within a Rails app folder.
-      :index_definition_path
+      :indices_path
 
     def self.delegated
       public_instance_methods - self.superclass.public_instance_methods - Singleton.public_instance_methods
@@ -53,7 +53,7 @@ module Chewy
       @root_strategy = :base
       @request_strategy = :atomic
       @use_after_commit_callbacks = true
-      @index_definition_path = 'app/chewy'
+      @indices_path = 'app/chewy'
     end
 
     def transport_logger= logger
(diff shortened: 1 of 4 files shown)


pyromaniac:
The only this here that concerns me - is the config entry name. `index_definition_path` is not quite clear. I would think about it. Everything else is awesome.

robacarp:
@pyromaniac I agree, but I couldn't come up with anything that seemed more obvious. I'm open to suggestions if you have them.

pyromaniac:
indices_path or indexes_path

pyromaniac:
I'm not sure, which one, I can't choose :) Actually, we've got both usages all over the gem. Probably it makes sense to pick one, indices, probably. So let's use `indices_path`.

robacarp:
@pyromaniac excellent! Updated

robacarp:
Sorry, It looks like I fat fingered that entire commit.

pyromaniac:
Awesome, thanks!

PR #415, opened 2016-08-16
Missing `script_fields` in the search delegate

dmitry:
After that fix readme example should run correctly:
https://github.com/toptal/chewy#script-fields


pyromaniac:
Perfect, thanks!

PR #420, opened 2016-09-04
Update travis to test against rails 5

pyromaniac:


Appraisals
```diff
@@ -1,4 +1,4 @@
-%w(3.2 4.0 4.1 4.2).each do |version|
+%w(3.2 4.2 5.0).each do |version|
```
krzysiek1507: `3.2`?

pyromaniac: well, yeah, we still support it

krzysiek1507: `3.2` gemfiles are missing from `gemfiles` part in `travis.yml`. Is this correct?

pyromaniac: https://github.com/toptal/chewy/pull/420/files#diff-354f30a63fb0907d4ad57269548329e3R28

They are in martig/include section.

Issue #422, opened 2016-09-06
Remove delete_by_query from journal


pyromaniac:
Since it is not supported by ES >= 2.0

PR #425, opened 2016-09-07
Remove delete by query plugin dependency from Chewy::Journal

sergey-kintsel:
for [this](https://github.com/toptal/chewy/issues/422)


pyromaniac:
#422

Issue #366, opened 2016-04-06
Why includes are removed from ActiveRecord scopes at index time ?


msievers:
I had some trouble using ActiveRecord scopes with chewy, till I discovered, the `includes` are removed from `scopes` and that was the cause, why my scopes *do* work on the console as expected, but using them to index the model *fails*.

The code I mean (introduced with `0.8.2`) is located here

https://github.com/toptal/chewy/blob/master/lib/chewy/type/adapter/active_record.rb#L45

I downgraded to version 0.8.1 and everything works fine.

Removing `includes` from scopes makes it a lot more complicated to use some more advanced scopes at index time. Therefore I would like to request to reconsider this decision.

pyromaniac:
Hey, sorry, I've just woke up :)

Can you provide an exact example when includes are required for id fetching?

msievers:
@pyromaniac Sure, it's even quite simple to construct one which fails.

Assume there are albums which may have records. There is a scope which only returns albums which actually have records.

```ruby
class Album < ApplicationRecord
  has_many :records

  scope :with_records, -> { includes(:records).where.not(records: { id: nil }) }
end

class Record < ApplicationRecord
  belongs_to :album
end
```

The corresponding index file should look like this

```ruby
class AlbumsIndex < Chewy::Index
  # this does not work, because the scope uses includes, which are removed
  define_type Album.with_records do
    field :title
  end
end
```

As soon as the `except(:includes)` statements are removed from `lib/chewy/type/adapter/active_record.rb` the scope can be used and `rake chewy:reset` correctly indexes 2 albums.

I created a basic rails app with these two models for you so that you can quickly reproduce this. Have a look at [msievers/chewy-includes-example](https://github.com/msievers/chewy-includes-example)

pyromaniac:
Hm, you are right, will remove it, or you can do it if you want and have some time.

msievers:
Because not removing the includes may have significant performance drawbacks in certain situations, I close this for now. See the attached PR for further informations.

PR #430, opened 2016-09-22
Revert removal of includes from scopes

msievers:
For details/motivation please see #366

### DIFF
--- a/lib/chewy/type/adapter/active_record.rb
+++ b/lib/chewy/type/adapter/active_record.rb
@@ -47,7 +47,7 @@ module Chewy
         end
 
         def pluck_ids(scope)
-          scope.except(:includes).uniq.pluck(target.primary_key.to_sym)
+          scope.distinct.pluck(target.primary_key.to_sym)
         end
 
         def scope_where_ids_in(scope, ids)
@@ -69,13 +69,25 @@ module Chewy
       end
 
       ActiveSupport.on_load(:active_record) do
-        if ::ActiveRecord::VERSION::MAJOR >= 5
-          module Rails5
+        if ::ActiveRecord::VERSION::MAJOR < 4
+          module Rails3
             def pluck_ids(scope)
-              scope.except(:includes).distinct.pluck(target.primary_key.to_sym)
+              # for activerecord < 4, pluck can't handle scopes with conditions using includes
+              scope.except(:includes).uniq.pluck(target.primary_key.to_sym)
             end
           end
-          Chewy::Type::Adapter::ActiveRecord.prepend(Rails5)
+          Chewy::Type::Adapter::ActiveRecord.prepend(Rails3)
+        elsif ::ActiveRecord::VERSION::MAJOR >= 4 && ::ActiveRecord::VERSION::MAJOR < 5
+          module Rails4
+            def pluck_ids(scope)
+              if scope.is_a?(::ActiveRecord::Associations::CollectionProxy)
+                scope.scope.distinct.pluck(target.primary_key.to_sym)
+              else
+                scope.distinct.pluck(target.primary_key.to_sym)
+              end
+            end
+          end
+          Chewy::Type::Adapter::ActiveRecord.prepend(Rails4)
         end
       end
     end
--- a/spec/chewy/type/import_spec.rb
+++ b/spec/chewy/type/import_spec.rb
@@ -43,6 +43,45 @@ describe Chewy::Type::Import do
       end
     end
 
+    describe 'using scopes with includes' do
+      if const_defined?('::ActiveRecord') && ::ActiveRecord::VERSION::MAJOR > 4
+        before do
+          stub_model(:country) do
+            has_many :cities
+
+            scope :with_cities, -> { includes(:cities).where.not(cities: { id: nil }) }
+          end
+
+          stub_model(:city) do
+            belongs_to :country
+          end
+
+          stub_index(:countries) do
+            define_type Country.with_cities do
+              field :name
+            end
+          end
+
+          Country.create(name: 'Russia', cities: City.create([{ name: 'Moscow' }, { name: 'Sevastopol' }]))
+          Country.create(name: 'China', cities: City.create([{ name: 'Beijing' }, { name: 'Shanghai' }]))
+          Country.create(name: 'Malta')
+        end
+
+        specify do
+          outer_payload = nil
+
+          ActiveSupport::Notifications.subscribe('import_objects.chewy') do |_name, _start, _finish, _id, payload|
+            outer_payload = payload
+          end
+
+          expect { CountriesIndex.import }.not_to raise_error
+          expect(outer_payload).to eq(type: CountriesIndex::Country, import: { index: 2 })
+        end
+      else
+        specify { skip 'only supported with ActiveRecord >= 4' }
+      end
+    end
+
     specify do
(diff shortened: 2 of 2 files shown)


spec/chewy/type/import_spec.rb
```diff
@@ -43,6 +43,45 @@
       end
     end
 
+    describe 'using scopes with includes' do
+      if const_defined?('::ActiveRecord') && ::ActiveRecord::VERSION::MAJOR > 4
```
pyromaniac: you can use `:active_record` tag for this context

lib/chewy/type/adapter/active_record.rb
```diff
@@ -69,13 +69,25 @@ def object_class
...
+              # for activerecord < 4, pluck can't handle scopes with conditions using includes
+              scope.except(:includes).uniq.pluck(target.primary_key.to_sym)
             end
           end
-          Chewy::Type::Adapter::ActiveRecord.prepend(Rails5)
+          Chewy::Type::Adapter::ActiveRecord.prepend(Rails3)
+        elsif ::ActiveRecord::VERSION::MAJOR >= 4 && ::ActiveRecord::VERSION::MAJOR < 5
+          module Rails4
+            def pluck_ids(scope)
+              if scope.is_a?(::ActiveRecord::Associations::CollectionProxy)
```
pyromaniac: What is it for? Can you provide an example of such a case?

msievers:
@pyromaniac Can you give this a review ?

I decided to leave the current implementation in place for ActiveRecord < 4, because its `pluck` cannot handle relation-based conditions anyway. For ActiveRecord 4 I handled a unconsistency between `ActiveRecord::Relation` and `ActiveRecord::Associations::CollectionProxy` and made the (sane) ActiveRecord > 4 code the default one.

msievers:
Unfortunatly the `:active_record` tag isn't enough here, because the scope does use `.where.not` which only works in ActiveRecord >= 4, so I have to guard this even for the ActiveRecord gemfiles. But if I don't check that `ActiveRecord` is present with this `const_defined?`, the spec will fail with `uninitialized constant ActiveRecord` for the non-AR gemfiles.

The `::ActiveRecord::Associations::CollectionProxy` is given (only) from the specs inside `spec/chewy/type/observe_spec.rb`, without knowing why in detail. I only noticed this because of (some) failing specs with the ActiveRecord 4 gemfiles.

Regarding performance I could do some imports with some real datasets to get a feeling, but I think in doubt, the user is free to use some "optimized" scopes for chewy if performance should really be an issue.

lib/chewy/type/adapter/active_record.rb
msievers: The `::ActiveRecord::Associations::CollectionProxy` is given (only) from the specs inside `spec/chewy/type/observe_spec.rb`, without knowing why in detail. I only noticed this because of (some) failing specs with the ActiveRecord 4 gemfiles.

spec/chewy/type/import_spec.rb
msievers: Unfortunatly the `:active_record` tag isn't enough here, because the scope does use `.where.not` which only works in ActiveRecord >= 4, so I have to guard this even for the ActiveRecord gemfiles. But if I don't check that `ActiveRecord` is present with this `const_defined?`, the spec will fail with `uninitialized constant ActiveRecord` for the non-AR gemfiles.

msievers:
Regarding performance I could do some imports with some real datasets to get a feeling, but I think in doubt, the user is free to use some "optimized" scopes for chewy if performance should really be an issue.

spec/chewy/type/import_spec.rb
msievers: Using something like this

```ruby
describe 'using scopes with includes', :active_record do
  if ::ActiveRecord::VERSION::MAJOR > 4
  ...
  end
end
```

does not work for the non-AR gemfiles and something like

```ruby
specify do
  skip unless ::ActiveRecord::VERSION::MAJOR > 4
  ...
end
```

does not work for the ar-3 gemfiles, because the `before` block is evaluated before one is able to skip-out and it will raise, because it cannot handle `where.not` correctly, which requires AR >= 4.

pr0d1r2:
👍  for measuring, compare and making decision

msievers:
Ok, just give me a 1-2 days. Will see to get some numbers quickly.

spec/chewy/type/import_spec.rb
conf: What about
```ruby
scope :with_cities, -> { includes(:cities).where('cities.id IS NOT NULL') }
```
?

msievers: Doesn't work because it wont construct the right (joined) query.

```ruby
Country.includes(:cities).where("cities.id IS NOT NULL").to_sql
=> "SELECT \"countries\".* FROM \"countries\" WHERE (cities.id IS NOT NULL)"
# will throw column not found for 'cities.id'
```

vs.

```ruby
Country.includes(:cities).where.not(cities: { id: nil }).to_sql
=> "SELECT \"countries\".\"id\" AS t0_r0, \"countries\".\"name\" AS t0_r1, \"countries\".\"country_code\" AS t0_r2, \"countries\".\"rating\" AS t0_r3, \"cities\".\"id\" AS t1_r0, \"cities\".\"country_id\" AS t1_r1, \"cities\".\"name\" AS t1_r2, \"cities\".\"rating\" AS t1_r3 FROM \"countries\" LEFT OUTER JOIN \"cities\" ON \"cities\".\"country_id\" = \"countries\".\"id\" WHERE (\"cities\".\"id\" IS NOT NULL)"
```
```

Comment (pyromaniac) [inline spec/chewy/type/import_spec.rb]: Yeah, in this case, you'll have to use `joins().includes()`

**Comment by msievers** [comment]
@pyromaniac It seems you are right regarding the performance implications when using pluck with a scope, which utilizes `includes`. I build a basic benchmark to get some numbers.

https://github.com/msievers/chewy/tree/benchmark/pluck_with_includes

While other operations are smart enough to detect that the included relation isn't needed at all, pluck doesn't. (even with the latest ActiveRecord 5).

```ruby
class Country < ActiveRecord::Base
  has_many :cities

  scope :with_includes, -> { includes(:cities) }
end

class City < ActiveRecord::Base
  belongs_to :country
end
```

```ruby
Benchmark.ips do |x|
  x.report('with includes') do
    Country.with_includes.pluck(:id)
  end

  x.report('without includes') do
    Country.pluck(:id)
  end

  x.compare!
end
```

```
Warming up --------------------------------------
       with includes     1.000  i/100ms
    without includes     6.000  i/100ms
       with includes      3.601  (± 0.0%) i/s -     18.000  in   5.000153s
    without includes     68.887  (± 1.5%) i/s -    348.000  in   5.053070s

Comparison:
    without includes:       68.9 i/s
       with includes:        3.6 i/s - 19.13x  slower
```

Dataset is `10000` countries, with `10` cities each.

The performance drop is hugh and I don't think this is reasonable. So I would close this, unless anybody has objections.

pyromaniac:
You are still able to use `includes().joins()` to bypass this behavior if you need. I've checked rails 5 and it behaves the same way as 4, no optimization there. Seems to be intended.

Anyway, I actually like how did you refactor rails 5 related module, it seems to be more accurate to me.

PR #432, opened 2016-09-26
Add support for namespaced indexes in journaling

### LINKED ISSUE #431: Problems with journaling and indexes in modules
Hi, we are having issues with `apply_changes_from` with journaling with indexes inside modules.

The way the index is serialized is using `_index_name` which calls `demodulize`. Then, when replaying the journal, `derive_type` is called to infer the index, but it is missing the module. `derive_type` does work with modules using the slash notation: `module/index#type`.

We are wondering if the fix should be in `_index_name` or in the way the index is serialized in the journal.

Any thoughts?
P.S. I did want to create a spec to reproduce the issue, but I can't seem to run the specs on `master`. `bundle` doesn't work: 

You have requested:
  listen ~> 3.0.0

The bundle currently has listen locked at 2.10.0.
Try running `bundle update listen`

If you are updating multiple gems in your Gemfile at once,
try passing them all to `bundle update`

fpbouchard:
Fixes #431. Let me know what you think. Thanks!

### DIFF
--- a/lib/chewy/index.rb
+++ b/lib/chewy/index.rb
@@ -42,6 +42,10 @@ module Chewy
       @_index_name
     end
 
+    def self.derivable_index_name
+      @_derivable_index_name ||= name.sub(/Index\Z/, '').underscore
+    end
+
     # Setups or returns pure Elasticsearch index name
     # without any prefixes/suffixes
     def self.default_prefix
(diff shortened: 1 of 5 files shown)


pyromaniac:
Wow. Exactly what I was thinking of. Almost. I was going to call it `reference` and join 2 journal columns in one with `#`,  but I this we can start with your patch. Going to review specs later and merge it. Awesome job, thanks!

pyromaniac:
Can you fix rubocop complains please?

fpbouchard:
Sorry about that.

pyromaniac:
Thanks!

PR #433, opened 2016-09-26
Use default indices_path path

dmitry:
Once I run `rake chewy:reset` it raised an error with:

```
TypeError: no implicit conversion of nil into String
```

Once I set `indices_path` in a chewy.yml file then this issue was solved.
Since this change now `indices_path` used by default, so you don't need to set it through chewy.yml.


pyromaniac:
Awesome, thanks!

PR #434, opened 2016-09-27
Only apply journal for current index when resetting

fpbouchard:
When resetting an index that has journaling, the current code actually replays all the journal entries, no matter which index has just been reset.

This PR only applies journal entries for the given index. I find my code a bit clunky (the new `query` method).

Any thoughts?

### DIFF
--- a/lib/chewy/journal.rb
+++ b/lib/chewy/journal.rb
@@ -48,10 +48,6 @@ module Chewy
       @records.any?
     end
 
-    def apply_changes_from(time)
-      Chewy::Journal.apply_changes_from(time, only: @index)
-    end
-
   private
 
     def identify(objects)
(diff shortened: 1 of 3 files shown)


pyromaniac:
Hey, awesome patch, will review it soon, thanks!

lib/chewy/journal.rb
```diff
@@ -123,21 +127,60 @@ def type_name
...
-        {
-          query: {
-            filtered: {
-              query_type => {
-                range: {
-                  created_at: {
-                    comparator => time.to_i
+      def query(time, comparator, index = nil, use_filter = true)
+        filter_query =
+          if use_filter
```
sergey-kintsel: maybe it makes sense to move some logic to separate methods? as for me, there are too many `if-else` statements in this method. Also, I would move this part to a separate method instead of just assigning it to `filter_query`

spec/chewy/journal_spec.rb
```diff
@@ -169,5 +169,52 @@ def timestamp(time)
...
+          expect(CountryIndex.all.to_a.length).to eq 2
+
+          # simulate lost data
+          Chewy.client.delete(index: 'city', type: 'city', id: 1, refresh: true)
+          Chewy.client.delete(index: 'country', type: 'country', id: 1, refresh: true)
+          expect(CityIndex.all.to_a.length).to eq 1
+          expect(CountryIndex.all.to_a.length).to eq 1
+
+          # Replay on specific index
+          Chewy::Journal.new(CityIndex).apply_changes_from(time)
```
sergey-kintsel: yeah, I would change the interface, to not pass an index name to the journal instance. I think it doesn't make sense. I would just pass it as `:only` parameter to`apply_changes_from` class method. Like there: `Chewy::Journal.apply_changes_from(time, only: [CityIndex])`

fpbouchard:
Updated it, let me know what you guys think.

pyromaniac:
Awecome, thanks!

PR #436, opened 2016-09-30
Fixed indices_path config setting

geraldke:
This bug seems to have been introduced by https://github.com/toptal/chewy/pull/433.

The `indices_path` setting from config will always be overwritten by `indices_path` which is `app/chewy`.

`configuration[:indices_path] = indices_path if indices_path`

This prevents `indices_path` setting from config from being overwritten by indices_path if the setting is not nil.

### DIFF
--- a/lib/chewy/config.rb
+++ b/lib/chewy/config.rb
@@ -112,7 +112,7 @@ module Chewy
     def configuration
       yaml_settings.merge(settings.deep_symbolize_keys).tap do |configuration|
         configuration[:logger] = transport_logger if transport_logger
-        configuration[:indices_path] = indices_path if indices_path
+        configuration[:indices_path] ||= indices_path if indices_path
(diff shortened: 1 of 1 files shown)


dsabados:
:+1:

pyromaniac:
Wow, right, didn't notice this, sorry, but actually it deserver some specs, WDYT?

geraldke:
@pyromaniac Yes, definitely. I'll follow up with the specs.

dmitry:
Sorry guys, haven't noticed configuration merge before setting `indices_path` when it's defined.
Just to understand: why anyone would like to set `indices_path` inside a environment specific settings `chewy.yml` file?

geraldke:
In my case, I wanted the indices to live in a different directory along with other indices (e.g. sphinx).

I've added a spec on https://github.com/toptal/chewy/pull/439

PR #437, opened 2016-09-30
Add chewy as the default Sidekiq queue

xirukitepe:
I added the same name of  `queue` when `Sidekiq` strategy is being invoked. The `ActiveJob` and `Resque` contain default queue named `chewy`, so might be helpful to add to `Sidekiq` as well.


pyromaniac:
👍

pyromaniac:
Thanks!

PR #439, opened 2016-10-02
Added specs for indices_path setting on configuration

geraldke:
This is adding a spec for https://github.com/toptal/chewy/pull/436

This test makes sure that the `indices_path` config in settings is not overridden by the default `indices_path`.


pyromaniac:
Thanks!

PR #440, opened 2016-10-03
add able to use Proc for settings

parallel588:


pyromaniac:
Hey, can you provide a reasoning as well as a couple of specs here please?

parallel588:
Hi Arkadiy (@pyromaniac), I added test.
about case where it may be use:
if we keep  a synonyms/stopwords in db then in the settings we can use synonyms from db
for instance:
```
 class ProductsIndex < Chewy::Index
  settings proc {
    {
      analysis: {
        filter: {
          synonym: {
            type: 'synonym',
            ignore_case: true,
            synonyms: SearchSynonym.synonyms
          },
          
          custom_stop: {
            type: 'stop',
            stopwords: SearchStopWord.terms
          },
.....
```

and we can update settings after changes SearchSynonym.synonyms

```
   Chewy.client.indices.put_settings(
        index: productsIndex.index_name,
        body: {
          index: {
            analysis: {
              filter: {
                synonym: {
                  synonyms: SearchSynonym.synonyms
                }
              } } } }
      )

```

pyromaniac:
Interesting, I would like to merge this, but can you fix a DSL here please? I would prefer

```
settings do

end
```

And rebase. And I'll merge it right away, thanks for the great idea.

pyromaniac:
Never mind, I'll do it by myself later, thanks for the great idea!

PR #417, opened 2016-08-24
Add the `track_scores` option to the query

dmitry:
`_score` to be computed and tracked even when there are no `_score` in sort.

More here: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html

PR #443, opened 2016-10-06
Add preference param to Query

menglewis:
This PR enables the `preference` option on `Query` which allows the user to control the preference for determining which shard replicas are used for a search request.

See: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-preference.html


pyromaniac:
Hey, awesome patch, thanks, but can you do like here as well: https://github.com/toptal/chewy/pull/417/files#diff-9640c36137c8265e53d4b946666de2c8R13

pyromaniac:
Thanks!

PR #449, opened 2016-10-18
Support Multi Instance Setup through Proc settings and Dynamic Client Keys

paulccarey:
This allows a multi instance Chewy/ES setup by allowing Procs to be specified for setting values.
The Proc values will be determined at runtime and the ES client cached as normal.

To avoid client cacheing clashes between tenants we specify an instance_identifier key that should also be a Proc.
e.g:

``` ruby
Chewy.settings = {
    host:                Proc.new { elastic_host_for_tenant },
    instance_identifier: Proc.new {tenant_identifier } 
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ tmp
 .rvmrc
 _site
 .sass-cache
+.idea/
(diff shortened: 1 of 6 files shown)


dalthon:
@paulccarey thanks for your contribution, but we would prefer to move configurations away from global configuration and initialize each new client as an object with its own config

PR #463, opened 2016-12-14
Ensure journal is empty

sergey-kintsel:
Refactor `Chewy::Journal`
Add multiple retries to apply the journal

### DIFF
--- a/lib/chewy/fields/base.rb
+++ b/lib/chewy/fields/base.rb
@@ -51,11 +51,13 @@ module Chewy
           object.send(name)
         end
 
-        result = if result.respond_to?(:to_ary)
-          result.to_ary.map { |item| compose_children(item, *objects) }
-        else
-          compose_children(result, *objects)
-        end if children.present? && !multi_field?
+        if children.present? && !multi_field?
+          result = if result.respond_to?(:to_ary)
+            result.to_ary.map { |item| compose_children(item, *objects) }
+          else
+            compose_children(result, *objects)
+          end
+        end
 
         { name => result }
       end
--- a/lib/chewy/index.rb
+++ b/lib/chewy/index.rb
@@ -90,14 +90,13 @@ module Chewy
       type_class = Chewy.create_type(self, target, options, &block)
       self.type_hash = type_hash.merge(type_class.type_name => type_class)
 
-      unless respond_to?(type_class.type_name)
-        class_eval <<-METHOD, __FILE__, __LINE__ + 1
-          def self.#{type_class.type_name}
-            ActiveSupport::Deprecation.warn("`#{self}.#{type_class.type_name}` accessor is deprecated and will be removed soon. Use `#{type_class}` directly instead.")
-            type_hash['#{type_class.type_name}']
-          end
-        METHOD
-      end
+      return if respond_to?(type_class.type_name)
+      class_eval <<-METHOD, __FILE__, __LINE__ + 1
+        def self.#{type_class.type_name}
+          ActiveSupport::Deprecation.warn("`#{self}.#{type_class.type_name}` accessor is deprecated and will be removed soon. Use `#{type_class}` directly instead.")
+          type_hash['#{type_class.type_name}']
+        end
+      METHOD
     end
 
     # Types method has double usage.
--- a/lib/chewy/journal/clean.rb
+++ b/lib/chewy/journal/clean.rb
@@ -1,0 +1,24 @@
+module Chewy
+  class Journal
+    module Clean
+      DELETE_BATCH_SIZE = 10_000
+
+      def until(time)
+        query = Query.new(time, :lte, nil, false).to_h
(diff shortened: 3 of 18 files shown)


DNNX:
The code looks ok. I sent you a few small remarks in DM. Let's :shipit: after green build.

lib/chewy/fields/base.rb
```diff
@@ -51,11 +51,13 @@ def compose(object, *parent_objects)
           object.send(name)
         end
 
-        result = if result.respond_to?(:to_ary)
-          result.to_ary.map { |item| compose_children(item, *objects) }
-        else
-          compose_children(result, *objects)
-        end if children.present? && !multi_field?
+        if children.present? && !multi_field?
```
pr0d1r2: Please consider splitting whole `compose` method as my eyes are bleeding:

```ruby
def compose(object, *parent_objects)
  { name => compose_result(object, parent_objects) }
end

def compose_result(object, *parent_objects)
  objects = ([object] + parent_objects.flatten).uniq
  result = base_result(object, objects)
  return compose_result_with_children(result, objects) if children.present? && !multi_field?
  result
end

def base_result(object, objects)
  if value && value.is_a?(Proc)
    result_from_proc(object, objects)
  elsif object.is_a?(Hash)
    result_from_hash(object)
  else
    object.send(name)
  end
end

def result_from_proc(object, objects)
  if value.arity.zero?
    object.instance_exec(&value)
  elsif value.arity < 0
    value.call(*object)
  else
    value.call(*objects.first(value.arity))
  end
end

def result_from_hash(object)
  if object.key?(name)
    object[name]
  else
    object[name.to_s]
  end
end

def compose_result_with_children(result, objects)
  if result.respond_to?(:to_ary)
    result.to_ary.map { |item| compose_children(item, *objects) }
  else
    compose_children(result, *objects)
  end
end
```

lib/chewy/index.rb
```diff
@@ -90,14 +90,13 @@ def self.define_type(target, options = {}, &block)
...
-          def self.#{type_class.type_name}
-            ActiveSupport::Deprecation.warn("`#{self}.#{type_class.type_name}` accessor is deprecated and will be removed soon. Use `#{type_class}` directly instead.")
-            type_hash['#{type_class.type_name}']
-          end
-        METHOD
-      end
+      return if respond_to?(type_class.type_name)
+      class_eval <<-METHOD, __FILE__, __LINE__ + 1
+        def self.#{type_class.type_name}
+          ActiveSupport::Deprecation.warn("`#{self}.#{type_class.type_name}` accessor is deprecated and will be removed soon. Use `#{type_class}` directly instead.")
```
pr0d1r2: Please break this long line by using heredoc:

```ruby
ActiveSupport::Deprecation.warn(<<WARNING.sqish)
  `#{self}.#{type_class.type_name}` accessor is deprecated and will be removed soon.
  Use `#{type_class}` directly instead.
WARNING

```

lib/chewy/journal/clean.rb
```diff
@@ -0,0 +1,24 @@
+module Chewy
+  class Journal
+    module Clean
+      DELETE_BATCH_SIZE = 10_000
+
+      def until(time)
+        query = Query.new(time, :lte, nil, false).to_h
+        search_query = query.merge(fields: ['_id'], size: DELETE_BATCH_SIZE)
+        index_name = Journal.index_name
```
pr0d1r2: We can use delegation:

```ruby
included do
  delegate :index_name, to: Journal
end
```

spec/chewy/journal/apply_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+            default_import_options journal: true
+          end
+        end
+        stub_index('country') do
+          define_type Country do
+            default_import_options journal: true
+          end
+        end
+
+        Chewy.massacre
```
pr0d1r2: 😅

lib/chewy/index.rb
sergey-kintsel: it is not connected to my changes) this was changed to make Rubocop happy

lib/chewy/fields/base.rb
sergey-kintsel: it is not connected to my changes) this was changed to make Rubocop happy

lib/chewy/journal/clean.rb
sergey-kintsel: actually, I think that explicit calls are easier) longer, but easier)

pr0d1r2: OK

lib/chewy/fields/base.rb
pr0d1r2: Can you please prepare follow-up PR after merging this one?

lib/chewy/index.rb
pr0d1r2: Can you please prepare follow-up PR after merging this one?

PR #467, opened 2017-01-13
Bulk Indexing Optimization and Refresh on Async Strategies

eproulx-petalmd:
Hello Pyro,

First of all, have a great year 👏👏

After some reading on [ElasticSearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html), it seems that dynamically disabling `refresh_interval` and decreasing `number_of_replicas` to 0 should enhance bulk indexing on indices. If thought it could be a nice configuration setting when zero downtime index resetting. I called it `use_enhance_index_settings_while_resetting`

Also, I added another configuration setting to disable `refresh` on async strategies like `sidekiq`, `resque` and `active_jobs`.  If you choose an async strategy, probably that you don't need a near-real time searchable document. Well in our case, we don't so I thought it would be nice addition. 

Thanks


lib/chewy/index/actions.rb
```diff
@@ -163,7 +163,29 @@ def #{method} options = {}
         def reset!(suffix = nil, journal: false)
           if suffix.present? && (indexes = self.indexes).present?
             create! suffix, alias: false
-            result = import suffix: suffix, journal: journal
+
+            result =
+              if Chewy.use_enhance_index_settings_while_resetting
+                name = build_index_name(suffix: suffix)
+                client.indices.put_settings index: name, body: { index: { number_of_replicas: 0, refresh_interval: -1 } }
```
pyromaniac: Can we refactor this? Make `number_of_replicas` a separate method with the block or so. I mean, this method became way huge, right?

eproulx-petalmd: True, its huge, i'll refactor it

lib/chewy/index/actions.rb
```diff
@@ -177,6 +184,24 @@ def reset!(suffix = nil, journal: false)
...
+          index_settings = options.slice :number_of_replicas, :refresh_interval
+          index_settings = extract_index_settings :number_of_replicas, :refresh_interval if index_settings.empty?
+          index_settings[:refresh_interval] = '1s' unless index_settings.key?(:refresh_interval)
+
+          client.indices.put_settings index: name, body: { index: index_settings }
+        end
+
+        def extract_index_settings(*args)
+          return {} unless settings_hash.key?(:settings) || settings_hash[:settings].key?(:index)
+          settings_hash[:settings][:index].select { |k, _| args.include?(k) }
```
DNNX: `settings_hash[:settings][:index].slice(*args)` is slightly simpler.

eproulx-petalmd:
rubocop is now 0.47.0 but I locked it to 0.46.0 since new offenses were flagged

eproulx-petalmd:
A new PR would be better for the 0.47.0

pyromaniac:
Sure it is ok. Did you take a look on my comment about separate options? What do you think?

.rubocop.yml
```diff
@@ -32,3 +32,8 @@ Style/MultilineMethodCallIndentation:
 
 Style/MultilineOperationIndentation:
   EnforcedStyle: indented
+
+Metrics/BlockLength:
```
pyromaniac: 👍 Now we need this in every project :)

eproulx-petalmd: 👍

eproulx-petalmd:
Honestly, I don't know about separating options. When would put back the original number_of_replicas if it was separated ? I would say the doc is confusing about that.

What do you think about the force_merge after ?

> And, a force merge should be called:

`curl -XPOST 'http://localhost:9200/test/_forcemerge?max_num_segments=5'`

pyromaniac:
I mean, setting refresh interval should be separate from setting the number of replicas. Somebody can decide to use either one, but not both.

eproulx-petalmd:
I would change the config settings to reflect user's choice

```ruby
 :only_refresh
 :only_replicas
 :both
```

eproulx-petalmd:
No prob

eproulx-petalmd:
Well, looks like `delete_by_query` has been updated for ES 5.1. 

https://github.com/elastic/elasticsearch-ruby/commit/cbfd6ebfe55c057666cc6de67fe6d9fc2f117908

The bad news is under version 5.0.0 it's not working anymore.

lib/chewy/config.rb
```diff
@@ -40,7 +40,19 @@ class Config
...
+      # If setting not present, put back to default 1s
+      # https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html
+      :reset_disable_refresh_interval,
+
+      # Set number_of_replicas to 0 before reset and put the original value after
+      # https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html
+      :reset_no_replicas,
+
+      # Refresh or not when import async (sidekiq, resque, activejob)
+      :disable_refresh_async
```
pyromaniac: Hey, why did you add this?
I'm thinking about some options passed to the block: `Chewy.strategy(:sidekiq, refresh: false) { do_stuff }` Instead of the global option. However it would not affect anything, so I'm ok to accept it if you don't want to change your mind.

eproulx-petalmd: I think both ways could work. We could set the strategy globally for the async `strategies` or we could simply pass some options to the block to have a different behavior for some cases. What do you think ?

pyromaniac: Yeah, this makes sense. I'm just thinking of it being slightly excess. But yeah, we can keep both ways. Tell me when you think this PR is ready, I'll review it again.

eproulx-petalmd: ready to review sir 👍

pyromaniac:
Ok, let's try it out. Thanks for this functionality!

PR #470, opened 2017-01-18
Add option to print progress while an index imports

### LINKED ISSUE #469: Allow printing of import progress during `rake chewy:reset`
I often tweak my chewy indexes on fairly large (2.5m row) ActiveRecord database tables. Running `rake chewy:reset` works, but often takes more than an hour and there's no indication of progress while the imports are taking place.

Also, I'm also in an unreliable SSH session while this is taking place--the socket tends to hang up after such a long period of no activity. It'd be nice to show a progress bar, percentage complete, or an ETA while a long import is taking place.

Is this something that'd be easy to implement? I can take a look at doing it myself.

gkaemmer:
First time touching this gem--not sure if this is abstracted enough.

Fixes #469.

### DIFF
--- a/lib/chewy/index/actions.rb
+++ b/lib/chewy/index/actions.rb
@@ -160,10 +160,10 @@ module Chewy
         #
         #   UsersIndex.reset! Time.now.to_i, journal: true
         #
-        def reset!(suffix = nil, journal: false)
+        def reset!(suffix = nil, journal: false, progress: false)
           if suffix.present? && (indexes = self.indexes).present?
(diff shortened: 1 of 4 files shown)


pyromaniac:
Hey. Everything is here actually: https://github.com/toptal/chewy/blob/master/lib/chewy/rake_helper.rb

gkaemmer:
Hmm, any other tips? I've been trying to use ActiveSupport::Notifications but it looks like Chewy only instruments the import as a whole, not the individual batches.

PR #481, opened 2017-03-10
add able to use Proc for settings

parallel588:
Hi Arkadiy (@pyromaniac), I added test.
about case where it may be use:
if we keep  a synonyms/stopwords in db then in the settings we can use synonyms from db
for instance:
```
 class ProductsIndex < Chewy::Index
  settings proc {
    {
      analysis: {
        filter: {
          synonym: {
            type: 'synonym',
            ignore_case: true,
            synonyms: SearchSynonym.synonyms
          },
          
          custom_stop: {
            type: 'stop',
            stopwords: SearchStopWord.terms
          },
.....
```

and we can update settings after changes SearchSynonym.synonyms

```
   Chewy.client.indices.put_settings(
        index: productsIndex.index_name,
        body: {
          index: {
            analysis: {
              filter: {
                synonym: {
                  synonyms: SearchSynonym.synonyms
                }
              } } } }
      )

```

### DIFF
--- a/spec/chewy/index/settings_spec.rb
+++ b/spec/chewy/index/settings_spec.rb
@@ -33,11 +33,11 @@ describe Chewy::Index::Settings do
     end
 
     specify do
-      Synonyms = Class.new do
+      stub_const('Synonyms', Class.new do
         def self.synonyms
           ['kaftan => dress']
         end
-      end
+      end)
       expect(
         described_class.new do
           {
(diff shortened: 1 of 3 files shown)


parallel588:
Hi @pyromaniac , I mistakenly closed the previous PR#440
Could you pleas check it PR?

pyromaniac:
Ah, here you are!

spec/chewy/index/settings_spec.rb
```diff
@@ -32,6 +32,27 @@
         .to eq(settings: { number_of_nodes: 3, analysis: {} })
     end
 
+    specify do
+      Synonyms = Class.new do
```
pyromaniac: Can you use stub_const here instead please?

pyromaniac:
Would appreciate if you will do a small fix, and I'll merge it right away, thanks!

pyromaniac:
Great, thanks!

PR #483, opened 2017-03-14
Add ability to skip index existence checks on each import

sergey-kintsel:
It is configurable by `skip_index_creation_on_import` parameter in
`config/chewy.yml`.

### DIFF
--- a/lib/chewy.rb
+++ b/lib/chewy.rb
@@ -202,5 +202,24 @@ module Chewy
       Chewy::Repository.instance
     end
     delegate(*Chewy::Repository.delegated, to: :repository)
+
+    def create_indices
+      Chewy::Index.descendants.each(&:create)
+    end
+
+    def create_indices!
+      Chewy::Index.descendants.each(&:create!)
+    end
+
+    def eager_load!
+      return unless defined?(Chewy::Railtie)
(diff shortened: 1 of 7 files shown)


lib/chewy.rb
```diff
@@ -202,5 +202,20 @@ def repository
       Chewy::Repository.instance
     end
     delegate(*Chewy::Repository.delegated, to: :repository)
+
+    def create_indices
+      eager_load!
+      Chewy::Index.descendants.each(&:create)
```
pyromaniac: `create!` would be better, also, `create_indices!`

lib/chewy/type/import.rb
```diff
@@ -237,6 +237,11 @@ def fetch_indexed_objects(objects)
 
           indexed_objects
         end
+
+        def assure_index_existence(index_options)
+          return if Chewy.configuration[:strict_index_creation]
```
pyromaniac: can we invert it and call `skip_index_creation_on_import`?

sergey-kintsel: okay

lib/chewy.rb
sergey-kintsel: I'll add both just to be able to use what we need

PR #487, opened 2017-03-22
ES5 journaling support

pyromaniac:


lib/chewy/journal/query.rb
```diff
@@ -4,65 +4,32 @@ class Query
...
-                should: @indices.collect { |i| index_filter(i) }
-              }
+      def query
+        {
+          query: {
+            bool: {
+              filter: [
+                range_filter,
+                bool: {
+                  should: @indices.collect { |i| index_filter(i) }
```
sergey-kintsel: maybe it may be done so:
```ruby
{
  bool: {
    must: [
      {
        terms: @indices.map(&:derivable_index_name)
      }
    ]
  }
}`

PR #488, opened 2017-03-23
README: Fix YAML example in journaling manual

olleolleolle:
This PR fixes the documentation of the README YAML configuration to enable journaling.

The previous value would become the String `"true,"`, not the `TrueClass` value.


pyromaniac:
Thanks!

PR #489, opened 2017-03-27
Support kaminari 1.0

pyromaniac:


lib/chewy.rb
```diff
@@ -19,6 +19,27 @@
 
 require 'elasticsearch'
 
+def try_require(path)
+  require path
+rescue LoadError
+  nil
+end
+
+try_require 'kaminari'
```
barthez: [`try_require 'kaminari'`](https://github.com/kaminari/kaminari/blob/master/lib/kaminari.rb) will require `kaminari/core` and `kaminari/activerecord` as well. Can you remove this line?

pyromaniac: Nope, this is for previous versions compatibility

PR #493, opened 2017-04-07
ES5 request

pyromaniac:
New request DSL, supports ES2 and ES5, replaces legacy `Chewy::Query` DSL


Spazer9:
I've tried this (maybe i'm wrong) but it doesn't seem to work well with the new `bool` query on ES 5. as described here:
https://www.elastic.co/guide/en/elasticsearch/reference/5.0/query-dsl-filtered-query.html

Which results in a parsing error:
`Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":{"root_cause":[{"type":"parsing_exception","reason":"no [query] registered for [filtered]","line":1,"col":22}],"type":"parsing_exception","reason":"no [query] registered for [filtered]","line":1,"col":22},"status":400}`

pyromaniac:
It still uses the old one, was to configured for the new one, unfortunately, will do it soon

pyromaniac:
There are actually other problems, like it doesn't support pagination or result objects yet

lib/chewy/search/parameters.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      def initialize(initial = {})
+        @storages = Hash.new do |hash, name|
+          hash[name] = self.class.storages[name].new
+        end
+        initial = initial.map do |name, value|
+          storage_class = self.class.storages[name]
+          storage = value.is_a?(storage_class) ? value : storage_class.new(value)
+          [name, storage]
+        end.to_h
```
olleolleolle: Using `each_with_object`, it would look like:

```ruby
initial.each_with_object({}) do |(name, value), result|
  storage_class = self.class.storages[name]
  storage = value.is_a?(storage_class) ? value : storage_class.new(value)
  result[name] = storage
end
```

lib/chewy/search/parameters.rb
```diff
@@ -55,7 +55,7 @@ def merge(other)
 
       def render
         body = @storages.values.inject({}) do |result, storage|
-          storage.render_to(result)
+          result.merge!(storage.render || {})
```
barthez: Can't we make `Storage#render` always return hash?

pyromaniac: It will require a lot of additional logic in every storage class

lib/chewy/search/request.rb
```diff
@@ -56,12 +56,14 @@ def reorder(value, *values)
...
+      %i(source stored_fields).each do |name|
+        define_method name do |value, *values|
+          modify(name) { update(values.empty? ? value : [value, *values]) }
+        end
       end
 
-      def stored_fields(value, *values)
-        modify(:stored_fields) { update(values.empty? ? value : [value, *values]) }
+      def search_after(value, *values)
+        modify(:search_after) { replace(values.empty? ? value : [value, *values]) }
```
barthez: What about this note:
image
Can we set `:from` (offset) to 0 here?

lib/chewy/search/parameters/docvalue_fields.rb
```diff
@@ -0,0 +1,19 @@
...
+    class Parameters
+      class DocvalueFields < Value
+        def update(value)
+          @value |= normalize(value)
+        end
+
+      private
+
+        def normalize(value)
+          Array.wrap(value).flatten.reject(&:blank?).map(&:to_s)
```
pyromaniac: Flatten just in case lol. And I don't like sets too much.

lib/chewy/search/request.rb
pyromaniac: Yeah, I'm going to do it later, going to use `except(:offset)` for this when it will be implemented :)

lib/chewy/search/parameters.rb
```diff
@@ -1,34 +1,5 @@
...
-require 'chewy/search/parameters/suggest'
-require 'chewy/search/parameters/docvalue_fields'
-require 'chewy/search/parameters/indices_boost'
-require 'chewy/search/parameters/min_score'
-require 'chewy/search/parameters/search_after'
-require 'chewy/search/parameters/rescore'
-require 'chewy/search/parameters/highlight'
-require 'chewy/search/parameters/load'
+Dir.glob(File.join(File.dirname(__FILE__), 'parameters', 'concerns', '*.rb')).each { |f| require f }
+Dir.glob(File.join(File.dirname(__FILE__), 'parameters', '*.rb')).each { |f| require f }
```
olleolleolle: Code golf detail: `Dir.glob` takes a block, too:

```
2.4.0 :002 > Dir.glob('/tmp/*') { |f| puts f.inspect }
"/tmp/com.apple.launchd.fq5DKZBr0e"
"/tmp/com.apple.launchd.IAoY1mzGsR"
"/tmp/com.apple.launchd.PRh2xPWk1O"
"/tmp/CrashUpload-Y12go"
"/tmp/ct.shutdown"
"/tmp/cvcd"
"/tmp/dumps"
"/tmp/mongodb-27017.sock"
"/tmp/mysql.sock"
"/tmp/mysql.sock.lock"
"/tmp/textmate-501.sock"
```

pyromaniac: wow, nice, thanks

.rubocop_todo.yml
```diff
@@ -13,7 +13,7 @@ Metrics/AbcSize:
 # Offense count: 4
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 290
+  Max: 300
```
barthez: Just disable this cop.

lib/chewy/query/criteria.rb
```diff
@@ -134,9 +134,9 @@ def storages
         STORAGES.map { |storage| send(storage) }
       end
 
-      def initialize_clone(other)
+      def initialize_clone(origin)
         STORAGES.each do |storage|
-          value = other.send(storage)
+          value = origin.send(storage)
```
olleolleolle: Dumb question: does this need to be `send` or can it be `public_send`?

pyromaniac: Yeah, it probably can, but I didn't configure rubocop to alert me yet :) Old habit, you know

lib/chewy/search/request.rb
```diff
@@ -19,7 +20,7 @@ def self.delegated_methods
            timeout min_score source stored_fields search_after
            load preload script_fields suggest indices_boost
            rescore highlight total total_count total_entries
-           types delete_all)
+           types unlimited delete_all count exists? exist? find)
```
barthez: Haven't you removed `unlimited`?

pyromaniac: Yeah, right, thanks

lib/chewy/search/request.rb
```diff
@@ -22,19 +38,18 @@ class Request
...
-          timeout min_score source stored_fields search_after
-          load script_fields suggest indices_boost
-          rescore highlight total total_count total_entries
-          types delete_all count exists? exist? find
-          scroll_batches scroll_hits scroll_results scroll_objects
-        ]
-      end
-
+      # The class is initialized with the list of chewy indexes and/or
+      # types, which are later used for the request composition.
```
olleolleolle: Could this line be slightly more active? `for the request composition.` => `to compose requests`?

lib/chewy/search/request.rb
```diff
@@ -1,17 +1,33 @@
...
-require 'chewy/search/query_proxy'
-require 'chewy/search/parameters'
-require 'chewy/search/response'
-require 'chewy/search/loader'
-
 module Chewy
   module Search
+    # The main requset DSL class. Supports multiple indexes requests.
+    #
+    # @note The class tries to be as much immutable as possible,
```
olleolleolle: Doc nitpick: You can drop the word `much` here.

lib/chewy/search/request.rb
pyromaniac: I appreciate your input, thanks

olleolleolle: Just watching this titanic, gigantic enormous work unfold is a pleasure.

pyromaniac: It is almost done, at least I hope so.

lib/chewy/search/parameters/concerns/query_storage.rb
```diff
@@ -40,6 +77,12 @@ def merge!(other)
           self.and(other.value)
```
barthez: What about `merge!` method?

lib/chewy/search/parameters/filter.rb
```diff
@@ -3,6 +3,17 @@
 module Chewy
   module Search
     class Parameters
+      # This parameter storage doesn't have it's own parameter at the
```
DNNX: it's -> its

lib/chewy/search/parameters/concerns/query_storage.rb
pyromaniac: Will do definitely, but later

lib/chewy/search/parameters.rb
```diff
@@ -3,16 +3,30 @@
 
 module Chewy
   module Search
+    # This class is basically a compoung storage of the request
```
olleolleolle: Typo `compoung` => `compound` 

`incapsulates` => `encapsulates`

lib/chewy/search/parameters.rb
```diff
@@ -24,30 +38,56 @@ def initialize(initial = {})
         end
       end
 
+      # Compares storages by their values
```
olleolleolle: Missing period (`.`)

lib/chewy/search/parameters.rb
```diff
@@ -3,16 +3,30 @@
 
 module Chewy
   module Search
+    # This class is basically a compoung storage of the request
+    # parameter storages. It incapsulates some storage-collection-handling
+    # logic.
+    #
+    # @see Chewy::Search::Request#parameters
     class Parameters
+      # Deafult storage classes warehouse. It is probably possible to
```
barthez: > ~Deafault~ Default

> It is ~probably~ possible to

lib/chewy/search/parameters.rb
```diff
@@ -3,16 +3,30 @@
...
+      # add your own classes here if necessary, but I'm not sure it will work.
```
barthez: It is very reassuring when the gem owner says that 😉

lib/chewy/search/parameters.rb
```diff
@@ -3,16 +3,30 @@
...
         @storages ||= Hash.new do |hash, name|
           hash[name] = "Chewy::Search::Parameters::#{name.to_s.camelize}".constantize
         end
       end
 
+      # @return [{Symbol => Chewy::Query::Parameters::Value}]
       attr_accessor :storages
       delegate :[], :[]=, to: :storages
 
+      # Accepts a hash of initial values as basic subobjects or
```
barthez: What are "subobjects"? Do you mean instances of subclasses? That way it would be much clearer.

lib/chewy/search/parameters.rb
```diff
@@ -24,30 +38,56 @@ def initialize(initial = {})
...
+      # @param name [Symbol] parameter name
+      # @yield the block is executed in the cloned storage instance binding
+      # @return [Chewy::Query::Parameters::Value]
       def modify!(name, &block)
         @storages[name] = @storages[name].clone.tap do |s|
           s.instance_exec(&block)
         end
       end
 
+      # Removes specified storages from the storages hash.
```
barthez: I suggest using `given` instead of `specified`.

lib/chewy/search/parameters.rb
pyromaniac: No I really have no idea, would it work or not :)

lib/chewy/search/parameters.rb
pyromaniac: ¯\_(ツ)_/¯

lib/chewy/search/parameters.rb
barthez: Then, just say it won't. Be on safe side.

lib/chewy/search/parameters.rb
pyromaniac: okay

pyromaniac: Actually, storage names are given, so it is "removes storages with given names", not "removes given storages" because you are not giving storages, names only., which, in short, is equal to "removes specified storages"

lib/chewy/search/parameters.rb
pyromaniac: No way, I love to walk on the edge

lib/chewy/search/parameters/storage.rb
```diff
@@ -1,34 +1,81 @@
...
+        # @see Chewy::Search::Parameters#merge!
+        # @see Chewy::Search::Request#merge
+        # @param other [Chewy::Search::Parameters::Storage] other storage
+        # @return [Object] updated value
         def merge!(other)
           update!(other.value)
         end
 
+        # Basic parameter rendering logic, don't need to return anything
+        # if parameter doesn't requre rendering for the current value.
```
olleolleolle: Typo `requre` => `require`

lib/chewy/search/loader.rb
```diff
@@ -0,0 +1,80 @@
...
+
+      # Returns a {Chewy::Type} object for index name and type name passed. Caches
+      # the result for each pair to make lookup faster.
+      #
+      # @param index [String] index name
+      # @param type [String] type name
+      # @return [Chewy::Type]
+      # @raise [NoMethodError] when index was not found
+      # @raise [Chewy::UndefinedType] when index was found, but type was not
+      def derive_type(index, type)
```
sergey-kintsel: I think `NoMethodError` is not that good explanation. Maybe it is worth to add there some meaningful one instead?

pyromaniac: Yeah, I'm actually going to rework the deriving logic later, that's why I left it like this

lib/chewy/search/request.rb
```diff
@@ -804,7 +804,7 @@ def exists?
       #   @return [Array<Chewy::Type>] result documents
       def find(*ids)
         ids = ids.flatten(1).map(&:to_s)
```
olleolleolle: Could you use `ids.flat_map(&:to_s)` here? https://ruby-doc.org/core-2.4.1/Enumerable.html#method-i-flat_map

olleolleolle: Answer: no.

```
2.4.0 :002 > [[1], [2], [3, 4]].flat_map { |e| e.to_s }
[
    [0] "[1]",
    [1] "[2]",
    [2] "[3, 4]"
]
2.4.0 :003 > [[1], [2], [3, 4]].flatten(1).map { |e| e.to_s }
[
    [0] "1",
    [1] "2",
    [2] "3",
    [3] "4"
]

```

pyromaniac: Ok :)

spec/chewy/search/request_spec.rb
```diff
@@ -0,0 +1,556 @@
...
+      end
+    end
+  end
+
+  describe '#filter' do
+    specify { expect(subject.filter(match: {foo: 'bar'}).render[:body]).to include(query: {bool: {filter: {match: {foo: 'bar'}}}}) }
+    specify { expect(subject.filter { match foo: 'bar' }.render[:body]).to include(query: {bool: {filter: {match: {foo: 'bar'}}}}) }
+    specify do
+      expect(subject.filter(match: {foo: 'bar'}).filter { multi_match foo: 'bar' }.render[:body])
+        .to include(query: {bool: {filter: {bool: {must: [{match: {foo: 'bar'}}, {multi_match: {foo: 'bar'}}]}}}})
```
barthez: Wouldn't it be better if it would render:
```
query: {bool: {filter: [{match: {foo: 'bar'}}, {multi_match: {foo: 'bar'}}]}}
```

it would have exact same behavior. So, if there is a filter with `must` clause only (no matter how many clauses are there) should render like this (similar to 2 examples above)

pyromaniac: Right, we probably need to add an additional logic to filter, will do

lib/chewy/search/parameters/concerns/query_storage.rb
```diff
@@ -16,7 +16,83 @@ class Parameters
       # @see Chewy::Search::Parameters::Filter
       # @see Chewy::Search::Parameters::PostFilter
       module QueryStorage
-        DEFAULT = {must: [], should: [], must_not: []}.freeze
+        # Bool storage value object, incapsulates update and query
```
olleolleolle: Typo `incapsulates` => `encapsulates`

.travis.yml
```diff
@@ -7,6 +7,7 @@ jdk:
 rvm:
   - 2.2.5
   - 2.3.3
+  - 2.4.1
```
olleolleolle: Also: 2.2.7, 2.3.4 exist.

pyromaniac: 👍

pyromaniac:
Whatever, merging this shit.

milos-v:
@pyromaniac will you bump up the version anytime soon? 
I really need your code. tnx!

pyromaniac:
Yeah, will do my best. Doing latest improvements at the moment

Issue #492, opened 2017-04-04
Do you plan to support ruby 2.4 ?


rekero:
After update i have a strange mistake. I think it's connected to a new instance_exec method presented in ruby 2.4.0.
NoMethodError - undefined method `__render__' for true:TrueClass
Did you mean?  __send__:
/.rvm/gems/ruby-2.4.1/bundler/gems/chewy-540462b77e84/lib/chewy/query/filters.rb:223:in `__render__'
/.rvm/gems/ruby-2.4.1/bundler/gems/chewy-540462b77e84/lib/chewy/query.rb:692:in `filter'
so __result__ return me True instead of valid query(in previous versions of ruby)
I tested both chewy 0.8.4 and 0.9.0
I thought it was connected to https://bugs.ruby-lang.org/issues/13074 , but 2.4.1 didn't solve this problem.

pyromaniac:
This is interesting, we are definitely going to, will check this out, thanks.

pyromaniac:
Btw, it seems like your block returns `true`, so this is not on the Chewy side. Can you post the contents of that filter block? I probably will be able to spot the problem.

rekero:
I have a class(i tried to simplify it) https://gist.github.com/rekero/e215515433d1f38ad075a31e8d0911f9
$ rvm use 2.2.3
Using /.rvm/gems/ruby-2.2.3
$ rails c
[1] pry(main)> OrdersQuery.new.first
  OrdersIndex::Order Search (44.1ms) {:body=>{:query=>{:filtered=>{:query=>{:match_all=>{}}, :filter=>{:term=>{"user_id"=>1}}}}}, :index=>["orders"], :type=>["order"]}
nil
$ rvm use 2.4.0
Using /.rvm/gems/ruby-2.4.0
$ rails c
[1] pry(main)> OrdersQuery.new.first
NoMethodError: undefined method `__render__' for true:TrueClass
Did you mean?  __send__

pyromaniac:
This actually should not work even in 2.3 :)
`o { user_id } == user_id` - this thing have to be written as `user_id == o { user_id }`. `o { user_id }` here is an outer scope variable, it doesn't have `==` method redefined and returns boolean as planned. While `user_id` is an instance of the DSL class which have `==` method redefined to return result term query.

rekero:
wow, i didn't know that order would matter, thanks

pyromaniac:
So, according to #497 chewy supports 2.4 completely

PR #497, opened 2017-05-10
Test against Ruby 2.4.1

shaneog:
Test against Ruby 2.4.1 on TravisCI.

Ref #492

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ jdk:
 rvm:
   - 2.2.5
   - 2.3.3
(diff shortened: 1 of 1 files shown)


pyromaniac:
Yeah, this is actually my next goal after ES 5 support, but go on :)

shaneog:
@pyromaniac The tests all seem to be passing with Ruby 2.4.1, but I'm afraid I might be out of my depth to correct #492.

pyromaniac:
Already in master

PR #510, opened 2017-06-02
Fixes check on class of index.

kevingriffin:
Before, the class was checked to see if it was an instance
of the base class it inherits from.

### DIFF
--- a/lib/chewy/rake_helper.rb
+++ b/lib/chewy/rake_helper.rb
@@ -31,7 +31,7 @@ module Chewy
       end
 
       def normalize_index(index)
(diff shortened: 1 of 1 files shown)


pyromaniac:
Hm, how did we miss this? :( Did it break some rake task usages?

kevingriffin:
It did—I described the issue here: https://github.com/toptal/chewy/issues/511

:)

pyromaniac:
Ah, got it, so it didn't affect rake task calls, it was affecting direct method calls only.

pyromaniac:
Ok, merged, enjoy. And thanks!

PR #514, opened 2017-06-05
Repairs class check in rake helper.

kevingriffin:
The previous fix broke string index names.


pyromaniac:
Can you add a spec please? :)

kevingriffin:
I was thinking something similar—I'll take a look at that this afternoon. Sorry for the breakage!

pyromaniac:
I'm merging it since it was found out that it breaks our stuff as well, but will wait for specs :)

PR #521, opened 2017-06-14
Partial import support

pyromaniac:
Partial import is a feature allows to update only specified fields on import


lib/chewy/rspec/update_index.rb
```diff
@@ -250,4 +244,17 @@ def array_difference(first, second)
...
   end
+
+  class Collector
+    attr_reader :updates
+
+    def initialize
+      @updates = []
+    end
+
+    def perform(body)
```
sergey-kintsel: where is it used? and why does it return an empty hash?

pyromaniac: It is used to collect data only. It mimics under the `Chewy::Type::Import::Bulk` class, repeats its API and returns empty hash to show that everything is fine. It is some kind of a mock.

lib/chewy/type/import/bulk.rb
```diff
@@ -0,0 +1,76 @@
...
+          if @bulk_size
+            pieces = body.each_with_object(['']) do |piece, result|
+              operation, meta = piece.to_a.first
+              data = meta.delete(:data)
+              piece = [{operation => meta}, data].compact.map(&:to_json).join("\n")
+
+              if result.last.bytesize + piece.bytesize > @bulk_size
+                result.push(piece)
+              else
+                result[-1] = [result[-1], piece].reject(&:blank?).join("\n")
```
sergey-kintsel: result[-1] = result.last. please, don't use negative indexes 🙏

lib/chewy/rspec/update_index.rb
sergey-kintsel: Could you leave it as a comment there? because it is not clear from the declaration

lib/chewy/type/import/bulk.rb
```diff
@@ -0,0 +1,76 @@
...
+          raise ArgumentError, '`bulk_size` can\'t be less than 1 kilobyte' if @bulk_size && @bulk_size <= 0
+        end
+
+        # Performs a bulk request with the passed body, returns empty
+        # array if everything is fine and array filled with errored
+        # document entries if something went wrong.
+        #
+        # @param body [Array<Hash>] a standard bulk request body
+        # @return [Array<Hash>] an array of bulk errors
+        def perform(body)
```
sergey-kintsel: according to the doc `body` parameter is an array of hashes. so, it would be better to name it `bodies` or maybe even better `bulk_pieces` or `bulk_parts`. WDYT?

lib/chewy/type/import/bulk.rb
```diff
@@ -0,0 +1,76 @@
...
+          return [] if body.blank?
```
sergey-kintsel: if `body` is an array, maybe explicitly `empty?` ?

pyromaniac: It can be anything actually, even nil

lib/chewy/type/import/bulk.rb
pyromaniac: nope, it is a body. It consists of documents

lib/chewy/type/import/bulk.rb
pyromaniac: well, I would prefer to be consistent here, don't see how .last is better in this case

lib/chewy/rspec/update_index.rb
pyromaniac: Yeah, will try to

lib/chewy/type/import/bulk.rb
sergey-kintsel: `.last` is clearer than negative indexes

pyromaniac: I think it is a personal opinion

sergey-kintsel: Of course

PR #524, opened 2017-06-21
Consistency checks implementation

pyromaniac:


lib/chewy/type/adapter/object.rb
```diff
@@ -54,6 +54,49 @@ def identify(collection)
           Array.wrap(collection)
         end
 
+        # For the object adapter this method tries to fetch :id and requested
+        # fields from the target `import_all_method` if defined. If not - it tries
+        # to call the target `pluck_method`, which is configurable and `pluck` by
+        # default. Basically, this `pluck_method` have to act exactly the same
```
mlen:
```
Basically, this `pluck_method` have to act exactly the same
```
↓
```
`pluck_method` has to act exactly the same
```

lib/chewy/type/adapter/object.rb
```diff
@@ -54,6 +54,49 @@ def identify(collection)
...
+        # way as the AR one. If none of the methods found - it simply returns
```
mlen: It returns an empty array when none of the methods are found.

lib/chewy/type/adapter/object.rb
```diff
@@ -54,6 +54,49 @@ def identify(collection)
           Array.wrap(collection)
         end
 
+        # For the object adapter this method tries to fetch :id and requested
+        # fields from the target `import_all_method` if defined. If not - it tries
```
mlen: `if defined` -> `when defined`. `If not -` -> `Otherwise`

lib/chewy/type/adapter/base.rb
```diff
@@ -36,6 +36,14 @@ def identify(_collection)
           raise NotImplementedError
         end
 
+        # Basically, performs the pluck method for the passed fileds on the
```
mlen: tyop in `fileds`, get rid of `Basically` :smile:

lib/chewy/type/mapping.rb
```diff
@@ -173,6 +175,15 @@ def mappings_hash
           root_object ? root_object.mappings_hash : {}
         end
 
+        # Does the type have incremental_update_field field defined? And is this
```
barthez: I would rather state it in simple sentences:
> Check whether a type has incremental_update_field set and that field is defined with a simple value.

Also, I'm still considering using `outdated_sync` instead of `increamental_update`. What do you think about that?

pyromaniac: Yeah, sure, I agree

lib/chewy/type/adapter/orm.rb
```diff
@@ -110,14 +110,15 @@ def load(ids, **options)
       private
 
         def import_objects(collection, options)
-          hash = Hash[identify(collection).zip(collection)]
+          collection_ids = identify(collection)
+          hash = Hash[collection_ids.map(&:to_s).zip(collection)]
```
barthez: Why we need hash with string keys?

pyromaniac: Ids exist in different formats, like mongo id or string or integer, string is a universal representation for all of them, I'm using it a lot.

PR #527, opened 2017-06-23
Add default field type setting

barthez:
Allow to sets default type for any field. By default, it is set to 'string'.

### DIFF
--- a/lib/chewy/stash.rb
+++ b/lib/chewy/stash.rb
@@ -8,7 +8,7 @@ module Chewy
     index_name 'chewy_stash'
 
     define_type :specification do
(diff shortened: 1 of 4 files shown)


lib/chewy/stash.rb
```diff
@@ -8,7 +8,7 @@ class Stash < Chewy::Index
     index_name 'chewy_stash'
 
     define_type :specification do
-      field :value, type: 'string', index: 'not_analyzed'
+      field :value, index: 'no'
```
pyromaniac: Are you sure this works for ES2?

barthez: Yup. I am.

PR #531, opened 2017-06-27
Elastic filters

averell23:
Filtered queries do no longer work in Elasticsearch 5 as the DSL changed
from what is encoded in chewy.

I don't assume that will be accepted as is, but I wanted to submit it 
as a proof of concept. I'm still testing if it works as expected.

Also, the "new" yaml loader does safe load and not accepts yaml references,
which is a pain.

### DIFF
--- a/lib/chewy/config.rb
+++ b/lib/chewy/config.rb
@@ -149,7 +149,7 @@ module Chewy
 
           if File.exist?(file)
             yaml = ERB.new(File.read(file)).result
-            hash = YAML.safe_load(yaml)
+            hash = YAML.safe_load(yaml, [], [], true)
             hash[Rails.env].try(:deep_symbolize_keys) if hash
           end
         end || {}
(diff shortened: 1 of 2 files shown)


averell23:
I see the yaml thing was already fixed ;-)

averell23:
So I leave this here for other people to maybe pick it up...

pyromaniac:
Hey, use the latest master and the new Request DSL

pyromaniac:
I'm not planning to do anything with the old DLS, it will be removed in the next versions

averell23:
I'm using the current master and was under the impression that I **do** use the new DSL. At least according to the README... Is there more than one filter method?

pyromaniac:
Well, you've fixed the old one - `Chewy::Query`, new one is completely separate and the entry point is in `Chewy::Search::Request`

pyromaniac:
Can I close this?

averell23:
Yes. Don't know, I think some part of our code was hardwired to Chewy::Query so this cropped up. It works now, I couldn't even reproduce it later...

pyromaniac:
Cool, thanks!

PR #532, opened 2017-06-27
Add shoryuken async strategy

josephchoe:
I added a new strategy for AWS SQS-based queues using the shoryuken gem.

### DIFF
--- a/lib/chewy/strategy/shoryuken.rb
+++ b/lib/chewy/strategy/shoryuken.rb
@@ -1,0 +1,39 @@
+module Chewy
+  class Strategy
+    # The strategy works the same way as atomic, but performs
+    # async index update driven by shoryuken
+    #
+    #   Chewy.strategy(:shoryuken) do
+    #     User.all.map(&:save) # Does nothing here
+    #     Post.all.map(&:save) # And here
+    #     # It imports all the changed users and posts right here
+    #   end
+    #
+    class Shoryuken < Atomic
+      class Worker
+        include ::Shoryuken::Worker
+
+        shoryuken_options auto_delete: true,
+                          body_parser: :json
+
+        def perform(sqs_msg, body, options = {})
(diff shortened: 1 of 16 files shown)


pyromaniac:
Cool, can you also fix the readme and the changelog please?

josephchoe:
Done.

pyromaniac:
ok, can you add the changelog entry on top, please? Also, did you test it manually? Does it work? If so, I'll merge it as soon as specs pass

josephchoe:
Oops, sorry.

josephchoe:
The specs pass, and I've tested it manually. But I think I've noticed a problem in copying the worker and specs from the sidekiq strategy to shoryuken. Shoryuken only passes in sqs_msg and body into the worker and not options, which would be a problem for passing in `suffix`.

https://github.com/phstc/shoryuken/blob/master/lib/shoryuken/processor.rb#L17

pyromaniac:
can we use `Shoryuken::Worker.perform_async({index: type, ids: ids, options: options}, queue: shoryuken_queue)` ?

lib/chewy/strategy/shoryuken.rb
```diff
@@ -0,0 +1,39 @@
...
+        def perform(_sqs_msg, body, options = {})
+          options[:refresh] = !Chewy.disable_refresh_async if Chewy.disable_refresh_async
+          body['index'].constantize.import!(body['ids'], options)
+        end
+      end
+
+      def leave
+        @stash.each do |type, ids|
+          next if ids.empty?
+          Shoryuken::Worker.perform_async({index: type, ids: ids}, queue: shoryuken_queue)
```
pyromaniac: Also, how about to use `{type: type` ?

josephchoe: Will do.

josephchoe:
Made the requested changes. Let me know if there's anything else that needs to be addressed.

pyromaniac:
Cool, merged, thanks!

PR #534, opened 2017-07-04
Fix deprication warnings for ES5

mpospelov:
1. use `text` type instead of `string` 
2. use `keyword` type instead of `not_analyzed` index type of `string` 
3. pass `match_all` query in case when plain index is passed to `delete_all`
4. replace `index: 'no'` with `index: false`

I found this WARNs in elastic search during development and decided to fix them

```
[2017-07-04T20:42:16,904][WARN ][o.e.d.i.m.TypeParsers    ] Expected a boolean [true/false] for property [index] but got [no]
[2017-07-04T20:44:15,342][WARN ][o.e.d.a.b.b.DeleteByQueryRequest] A request to _delete_by_query without an explicit query is deprecated. Specify a query explicitly instead.
[2017-07-04T20:35:52,280][WARN ][o.e.d.i.m.StringFieldMapper$TypeParser] The [string] field is deprecated, please use [text] or [keyword] instead on [ar_name]
```


pyromaniac:
Hey, there is a problem with this PR, we want to support both, ES2 and 5. I was going to provide fixes to the journal next week. But I'll take your findings in consideration, kudos for the documentation fixes!

pyromaniac:
Can you check the latest master please, we probably don't need this patch, thanks!

PR #536, opened 2017-07-05
Optimize and parallelize outdated calculation

pyromaniac:


lib/chewy/type/syncer.rb
```diff
@@ -19,10 +19,79 @@ class Type
...
+      # milliseconds, so ES stored time with the seconds precision.
+      if ActiveSupport::VERSION::STRING < '4.1.0'
+        def self.dates_equal(one, two)
+          one.to_i == two.to_i
+        end
+      end
 
       # @param type [Chewy::Type] chewy type
-      def initialize(type)
+      def initialize(type, parallel: 4)
```
barthez: What about making this `4` a Chewy configuration option?

pyromaniac: it will be parallel: nil and will be passed from the rake helper, I just didn't implement the interface yet

pyromaniac: Exactly the same way it is done for import

PR #537, opened 2017-07-07
Docs improvement

pyromaniac:


README.md
```diff
@@ -917,39 +921,91 @@ end
 
 ### Rake tasks
 
-#### `chewy:update` and `chewy:reset`
+For a Rails application, some index-maintaining rake tasks are defined.
 
-Inside the Rails application, some index-maintaining rake tasks are defined.
+#### `chewy:reset`
+
+Rerforms zero-downtime reindexing as described [here](https://www.elastic.co/blog/changing-mapping-with-zero-downtime). So the rake task creates a new index with uniq suffix and then simply aliases it to the common index name. The previous index is deleted afterwards (see `Chewy::Index.reset!` for more details).
```
mlen: Performs?

pyromaniac: Yup :)

README.md
```diff
@@ -917,39 +921,91 @@ end
...
 
 ```bash
 rake chewy:reset # resets all the existing indices, declared in app/chewy
 rake chewy:reset[users] # resets UsersIndex only
-rake chewy:reset[users,projects] # resets UsersIndex and ProjectsIndex
-rake chewy:reset[-users,projects] # resets every index in application except specified ones
+rake chewy:reset[users,places] # resets UsersIndex and PlacesIndex
+rake chewy:reset[-users,places] # resets every index in application except specified ones
```
mlen: It's not very obvious for me: does it skip both users and places? I'd suggest to use `[-users,-places]` syntax then

pyromaniac: It doesn't make sense actually, you can't include something and exclude something at the same time, such an action can be reduced to the whitelist only, so it doesn't make sense to have the minus-minus syntax.

pyromaniac: Can you propose the rephrase to make it more clear?

README.md
```diff
@@ -917,39 +921,91 @@ end
...
+```bash
+rake chewy:upgrade # upgrades all the existing indices, declared in app/chewy
+rake chewy:upgrade[users] # upgrades UsersIndex only
+rake chewy:upgrade[users,places] # upgrades UsersIndex and PlacesIndex
+rake chewy:upgrade[-users,places] # upgrades every index in application except specified ones
+```
+
+#### `chewy:update`
+
+Doesn't create indexes, it simply imports everything to the existing ones and fails if the index was not created before.
```
barthez: > It doesn't
I'm not sure you can skip object easily in English…

README.md
```diff
@@ -917,39 +921,91 @@ end
...
 
-This rake task is especially useful during the production deploy. Currently it executes selective reset, this means that an index will be reset only if the index specification (settings or mappings) has been changed, otherwise the reset of this index will be skipped.
+Arguments are similar to the `chewy:update` ones. It is possible to specify a particular type or a whole index.
 
-Obviously at the first run it will reset everything because it needs to lock all the index specifications in the `Chewy::Stash`.
+See [Chewy::Type::Syncer](lib/chewy/type/syncer.rb) for more details.
 
-See [Chewy::Stash](lib/chewy/stash.rb) and [Chewy::Index::Specification](lib/chewy/index/specification.rb) for more details.
+```bash
+rake chewy:sync # synchronizes all the existing indices, declared in app/chewy
```
barthez: The comment is not exactly true. It will sync all indexes that inherit from Chewy::Type. You can setup different chewy index definition location.

README.md
pyromaniac: Fine to me :) But fill fix it.

README.md
mlen: I don't have a better idea, sorry. I just don't like this syntax, but it makes sense that you can't specify both exclusions and inclusions at the same time

README.md
```diff
@@ -917,39 +921,89 @@ end
...
 rake chewy:reset[users] # resets UsersIndex only
-rake chewy:reset[users,projects] # resets UsersIndex and ProjectsIndex
-rake chewy:reset[-users,projects] # resets every index in application except specified ones
+rake chewy:reset[users,places] # resets UsersIndex and PlacesIndex
+rake chewy:reset[-users,places] # resets every index in application except specified ones
+```
+
+#### `chewy:upgrade`
+
+Performs reset exactly the same way as `chewy:reset` does, but only if the index specification (setting or mapping) have been changed.
```
mlen: `but only when`
`was changed`

README.md
```diff
@@ -917,39 +921,89 @@ end
...
 ```
 
-`rake chewy:reset` performs zero-downtime reindexing as described [here](https://www.elastic.co/blog/changing-mapping-with-zero-downtime). So basically rake task creates a new index with uniq suffix and then simply aliases it to the common index name. The previous index is deleted afterwards (see `Chewy::Index.reset!` for more details).
+#### `chewy:update`
 
-#### `chewy:deploy`
+It doesn't create indexes, it simply imports everything to the existing ones and fails if the index was not created before.
 
-This rake task is especially useful during the production deploy. Currently it executes selective reset, this means that an index will be reset only if the index specification (settings or mappings) has been changed, otherwise the reset of this index will be skipped.
+Unlike `reset` or `upgrade` tasks, it is possible to pass type references to update the particular type. In index name is passed without the type specified, it will update all the types defined for this index.
```
mlen: `all types` I think the article is not needed in plural

README.md
```diff
@@ -917,39 +921,89 @@ end
...
 
-Obviously at the first run it will reset everything because it needs to lock all the index specifications in the `Chewy::Stash`.
+```bash
+rake chewy:update # updates all the existing indices
+rake chewy:update[users] # updates UsersIndex only
+rake chewy:update[users,places#city] # updates the whole UsersIndex and PlacesIndex::City type
+rake chewy:update[-users,places#city] # updates every index in application except all the types defined in UsersIndex and the rest of the types defined in PlacesIndex
```
mlen: `in the application`, `except every type defined`?

README.md
```diff
@@ -917,39 +921,89 @@ end
...
+```bash
+rake chewy:sync # synchronizes all the existing indices
+rake chewy:sync[users] # synchronizes UsersIndex only
+rake chewy:sync[users,places#city] # synchronizes the whole UsersIndex and PlacesIndex::City type
+rake chewy:sync[-users,places#city] # synchronizes every index in application except all the types defined in UsersIndex and the rest of the types defined in PlacesIndex
+```
+
+#### `chewy:deploy`
+
+This rake task is especially useful during the production deploy. It is a combination of `chewy:upgrade` and `chewy:sync` and the latter one is called only for the indexes that were not reset during the first stage.
```
mlen: `and the latter is called only for these indices that were not reset during the first stage` (without one)

README.md
```diff
@@ -917,39 +921,89 @@ end
...
+
+Right now the approach is that if some data had been updated, but index definition had not been changed no changes satisfying the synchronization algorithm were done, it would be much faster to perform manual partial index update inside data migrations or even manually after the deploy.
 
 Also, there is always full reset alternative with `rake chewy:reset`.
 
+##### Parallelizing rake tasks
+
+Every task described above has its own parallel version. Every parallel rake task takes the number for processes for execution as the first argument and the rest of the arguments are exactly the same as for the non-parallel task version.
+
+To use this task it is require to install [https://github.com/grosser/parallel](https://github.com/grosser/parallel) gem.
```
mlen: It is required to install ... gem to use this task

README.md
```diff
@@ -917,39 +921,89 @@ end
...
+
+If the number of processes is not specified - `parallel` gem tries to derive it from the system information.
```
mlen: `parallel` gem tries to automatically derive the number of processes to run when it is not explicitly defined

README.md
```diff
@@ -917,39 +921,89 @@ end
...
+rake chewy:update[-users,places#city] # updates every index in application except all the types defined in UsersIndex and the rest of the types defined in PlacesIndex
+```
 
-See [Chewy::Stash](lib/chewy/stash.rb) and [Chewy::Index::Specification](lib/chewy/index/specification.rb) for more details.
+#### `chewy:sync`
+
+Provides a way to synchronize outdated indexes with the source quickly and without doing a full reset.
 
-In the future, additional routines are planned during `chewy:deploy` execution. Like, additional fast or partial index updates to make sure everything is up-to-date as it was a full reset.
+Arguments are similar to the `chewy:update` ones. It is possible to specify a particular type or a whole index.
```
mlen: >arguments are similar to the ones taken by `chewy:update` task
maybe?

README.md
```diff
@@ -917,39 +921,89 @@ end
...
+rake chewy:sync[-users,places#city] # synchronizes every index in application except all the types defined in UsersIndex and the rest of the types defined in PlacesIndex
+```
+
+#### `chewy:deploy`
+
+This rake task is especially useful during the production deploy. It is a combination of `chewy:upgrade` and `chewy:sync` and the latter one is called only for the indexes that were not reset during the first stage.
+
+It is not possible to specify any particular types/indexes for this task as it doesn't make much sense.
+
+Right now the approach is that if some data had been updated, but index definition had not been changed no changes satisfying the synchronization algorithm were done, it would be much faster to perform manual partial index update inside data migrations or even manually after the deploy.
```
mlen: `index definition had not been changed no changes satisfying the synchronization algorithm were done` there's something wrong here

README.md
pyromaniac: sometimes it might be used, I'm pretty sure this is the case

PR #539, opened 2017-07-11
Make _Parallelizing rake tasks_ h4 (was: h5)

DNNX:
According to the screenshot, _Parallelizing rake tasks_ should be on the same level as previous items.

Having _Parallelizing rake tasks_ section as a subsection of `chewy:deploy` doesn't really make much sense.

image

Before:

image

After:

image


pyromaniac:
Cool, thanks!

Issue #552, opened 2017-08-04
undefined method `debug` 


niels-s:
With the latest version 0.10.0 I got a undefined method `debug` when booting my rails app.

The issue is introduced by this PR I believe: https://github.com/toptal/chewy/pull/543
I think we have a race condition here, since the `Chewy.logger` has been initialised yet because ActiveRecord isn't load. Because ActiveRecord `on_load` is used on [lib/chewy/railtie.rb#55](https://github.com/toptal/chewy/blob/master/lib/chewy/railtie.rb#L55), the time when the RequestStrategy is initialised, the Chewy.logger is still nil in my case.

Using Rails 5.1.3 on ruby 2.4.0

pyromaniac:
Oh, okay, thanks for the report

PR #553, opened 2017-08-04
Make sure Chewy logger is initialised

niels-s:
Since the purpose of #543 was to reduce the strategy logging I thought this was the best solution imho without loosing the logging information and with my limited knowledge of the project.

If people with more experience of this project have better idea, I'm open for suggestions 😛 

Possible fix for #552 

cc: @pyromaniac, @Borzik


pyromaniac:
Oh crap, I didn't notice you did it. I've got a better solution thought anyway, take a look, please: #557

PR #557, opened 2017-08-06
Better initial strategies logging

### LINKED ISSUE #552: undefined method `debug`

pyromaniac:
Fixes #552 
Instead of: #553

### DIFF
--- a/lib/chewy/railtie.rb
+++ b/lib/chewy/railtie.rb
@@ -7,7 +7,6 @@ module Chewy
     class RequestStrategy
       def initialize(app)
         @app = app
-        Chewy.logger.debug("Chewy strategies stack: [1] <- #{Chewy.request_strategy}")
       end
 
       def call(env)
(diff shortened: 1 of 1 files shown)


arturtr:
@pyromaniac this logic requires initialize `Chewy.logger` in initializer

I solve this problem by this in initalizer
```ruby
Chewy.logger = Logger.new(nil)
```

pyromaniac:
Is it with this PR included?

arturtr:
Yes. This line 
https://github.com/toptal/chewy/blob/master/lib/chewy/railtie.rb#L53
broke rails to start 

because `Chewy.logger` is `nil` by default

arturtr:
I catch this on run rails console

pyromaniac:
Ah, console, got it, thanks. Will fix it

PR #567, opened 2017-08-25
Wrapper: missed check in highlight method

heartfulbird:
Hello, it is very helpful library! Thank you so much)

It seems you missed check with your own method in wrapper highlight method.
I'm not sure is it good practice or not to use elasticsearch without query or use it with wildcard, but in  this case we need to check that _data  contains key "highlights" and _data['highlights'] contains key which is equal to attribute name. 
Moreover - you already have this method!)


pyromaniac:
Makes sense actually, thanks!

PR #542, opened 2017-07-19
Fix symbol splitting deprecation warning

ianmurrays:
Sequel throws a warning due to symbol splitting, and suggests using `Sequel.qualify`
to get rid of it. Symbol splitting is deprecated and will be removed in Sequel 5.

PR #577, opened 2017-09-15
Fix Sequel deprecation warnings

matchbookmac:
## Why
Sequel throws deprecation warnings when using its older symbol-splitting syntax (`:table__column`).
Additionally, `Model.table_name` in `sequel` can return a non-`String`/`Symbol` value.

## How
Uses `::Sequel.qualify` to join `table_name` and `column`.
Uses the `Sequel::SQL::QualifiedIdentifier` returned from `Chewy::Type::Adapter::Sequel#full_column_name` appropriately in the relevant `scope` filters.

### Notes
I see there is already a PR for this (#542), but it looks like the build isn't passing.

### DIFF
--- a/lib/chewy/type/adapter/sequel.rb
+++ b/lib/chewy/type/adapter/sequel.rb
@@ -34,7 +34,7 @@ module Chewy
         end
 
         def full_column_name(column)
-          "#{target.table_name}__#{column}".to_sym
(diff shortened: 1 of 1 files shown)


pyromaniac:
Awesome, thanks!

matchbookmac:
Happy to help!

PR #580, opened 2017-09-19
Disconnect sequel databases before forking in parallel

matchbookmac:
## Why
Much like `ActiveRecord`, `Sequel` cannot share a database connection between processes. Connections must be closed before forking a new process so that the new process will establish its own connection.

## How
Calls `#disconnect` on each connection `Sequel` holds. `Sequel` reconnects to the database before executing any query so there is no need to call `#reconnect` as there is with `ActiveRecord`.


matchbookmac:
Specs use sqlite as the test db. Doesn't play so nice with `#disconnect` since it's all in memory. Will try to find another way.

pyromaniac:
Ok, cool, just keep in mind that there are several places like this, please. Looking forward to the patch.

matchbookmac:
Noted, thanks for the tip!

PR #584, opened 2017-10-02
Add compatibility for ruby < 2.2.0

vorless:
Prior to ruby version 2.2.0 the `include` method was private. Use `send` here to maintain compatibility through older versions.


pyromaniac:
Ok, but 2.1 is not supported and tests are not running for it.

PR #585, opened 2017-10-03
Refactor root field to be less restrictive

pyromaniac:


lib/chewy/fields/root.rb
```diff
@@ -6,15 +6,23 @@ class Root < Chewy::Fields::Base
       attr_reader :parent
       attr_reader :parent_id
 
-      def initialize(*args)
-        super(*args)
+      def initialize(*_)
```
mlen: When you don't use it, it may be `def initialize(*)` as well

lib/chewy/type/mapping.rb
```diff
@@ -33,9 +33,11 @@ module ClassMethods
         #     end
         #   end
         #
-        def root(options = {}, &block)
-          raise 'Root is already defined' if root_object
-          build_root(options, &block)
+        def root(options = {})
```
barthez: You are not supporting Ruby < 2.0, consider using `**options` instead. You will gain that keywords arguments splat accepts only hashes with symbols as keys, no need to convert them down the road.

lib/chewy/fields/root.rb
```diff
@@ -6,15 +6,23 @@ class Root < Chewy::Fields::Base
...
+
+      def update_options!(options)
+        options.deep_symbolize_keys!
+        @id = options.delete(:id) if options.key?(:id)
+        @id = options.delete(:_id) if options.key?(:_id)
+        @parent = options.delete(:parent) if options.key?(:parent)
+        @parent = options.delete(:_parent) if options.key?(:_parent)
+        @parent_id = options.delete(:parent_id) if options.key?(:parent_id)
+        options.delete(:type)
+        @options.merge!(options)
```
mlen: Maybe replace the deletions with `fetch` and `except`? Activesupport is already a dependency.
`options.fetch(:_id, options[:id])`, etc and `@options.merge!(options.except(:parent_id, :parent, :_parent, :id, :_id))`

lib/chewy/type/mapping.rb
```diff
@@ -165,46 +165,41 @@ def agg(name, &block)
         #   template template42: {match: 'hello*', mapping: {type: 'object'}} # or even pass a template as is
         #
         def template(*args)
-          build_root.dynamic_template(*args)
+          root.dynamic_template(*args)
```
mlen: Can't `root_object` be used directly when it is not updated here?

lib/chewy/type/wrapper.rb
```diff
@@ -70,7 +70,7 @@ def highlight_name(method)
       end
 
       def attribute_defined?(attribute)
-        self.class.root_object && self.class.root_object.children.find { |a| a.name.to_s == attribute }.present?
+        self.class.root && self.class.root.children.find { |a| a.name.to_s == attribute }.present?
```
mlen: The same here, when root object is not updated, there should be no need to run the code that is essentially a noop

lib/chewy/type/mapping.rb
pyromaniac: Don't like the idea, there should be some DSL. Are you proposing to get rid of this method?

lib/chewy/type/wrapper.rb
pyromaniac: I want less exceptions tbh

PR #590, opened 2017-10-26
Always ignore root in as_json to ignore include_root_in_json in AR

Strnadj:
We're porting the current Chewy version to ActiveSupport 3.2. (Official support was dropped, don't really know why - you just need to port pluck and stringify_keys from Rails 4 and that's almost all..).

But, in Rails 4 came option `ActiveRecord::Base.include_root_in_json = true`, with this configuration all `as_json` return root key. (Like it was default in Rails 3).


pyromaniac:
Cool, thanks!

PR #593, opened 2017-10-27
Fixes logic error in index settings

yahooguntu:
There's a logic error in the code that extracts an index's settings from the settings hash.

Only occurs if you have these two settings enabled (and probably any
other config option that saves/restores index settings):
 * reset_disable_refresh_interval
 * reset_no_replicas

### DIFF
--- a/spec/chewy/index/actions_spec.rb
+++ b/spec/chewy/index/actions_spec.rb
@@ -429,6 +429,12 @@ describe Chewy::Index::Actions do
             expect(CitiesIndex).to receive(:import).with(suffix: suffix, journal: false, refresh: false).and_call_original
             expect(CitiesIndex.reset!(suffix)).to eq(true)
           end
+
+          specify 'uses empty index settings if not defined' do
+            allow(Chewy).to receive(:wait_for_status).and_return(nil)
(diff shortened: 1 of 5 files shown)


pyromaniac:
Hey, great finding, can you add a spec for the case, please?

yahooguntu:
@pyromaniac How's that look?

spec/chewy/index/actions_spec.rb
```diff
@@ -429,6 +429,12 @@
             expect(CitiesIndex).to receive(:import).with(suffix: suffix, journal: false, refresh: false).and_call_original
             expect(CitiesIndex.reset!(suffix)).to eq(true)
           end
+
+          specify 'uses empty index settings if not defined' do
+            allow(Chewy).to receive(:wait_for_status).and_return(nil)
+            allow(CitiesIndex).to receive(:settings_hash).and_return({})
```
pyromaniac: Did you check if it fails without your fix?

yahooguntu: Yeah, the root cause was because of spec/spec_helper.rb:27 ; if the :index option isn't set, then the tests fail because Index#index_settings tries to access it.

pyromaniac:
Thanks for the contribution!

Issue #595, opened 2017-10-28
DocValuesField \"value\" is too large, must be <= 32766


kevgrig:
Hi, I'm calling `reset!` on my index and it's throwing the following exception. Chewy is not providing details on how to investigate this:

```
$ bin/rails console
[...]
irb(main):001:0> UserIndex.reset!
[...]
D, [2017-10-28T13:58:24.834739 #7851] DEBUG -- :   Chewy::Stash::Specification Import (8.5ms) {:index=>1}
Chewy::ImportFailed: Import failed for `Chewy::Stash::Specification` with:
    Index errors:
      `{"type"=>"illegal_argument_exception", "reason"=>"DocValuesField \"value\" is too large, must be <= 32766"}`
        on 1 documents: ["user"]

        from chewy (0.10.1) lib/chewy/type/import.rb:88:in `import!'
        from chewy (0.10.1) lib/chewy/index/specification.rb:25:in `lock!'
        from chewy (0.10.1) lib/chewy/index/actions.rb:199:in `reset!'
        from (irb):1
        from railties (5.1.4) lib/rails/commands/console/console_command.rb:62:in `start'
        from railties (5.1.4) lib/rails/commands/console/console_command.rb:17:in `start'
        from railties (5.1.4) lib/rails/commands/console/console_command.rb:97:in `perform'
        from thor (0.20.0) lib/thor/command.rb:27:in `run'
        from thor (0.20.0) lib/thor/invocation.rb:126:in `invoke_command'
        from thor (0.20.0) lib/thor.rb:387:in `dispatch'
        from railties (5.1.4) lib/rails/command/base.rb:63:in `perform'
        from railties (5.1.4) lib/rails/command.rb:44:in `invoke'
        from railties (5.1.4) lib/rails/commands.rb:16:in `<top (required)>'
        from bin/rails:4:in `require'
        from bin/rails:4:in `<main>'
```

Nothing in my models with "value":

```
$ grep -c "\"value\"" db/schema.rb 
0
```

Index definition: https://github.com/myplaceonline/myplaceonline_rails/blob/master/app/lib/user_index.rb

pyromaniac:
Hm, interesting, you've got a huge index definition, so it can't be stored in `Chewy::Stash`. Try to fix the definition of `Chewy::Stash`, this field https://github.com/toptal/chewy/blob/master/lib/chewy/stash.rb#L13 needs an additional option `doc_values: false`

kevgrig:
[...]

PR #596, opened 2017-10-29
Issue #595: Disable doc_values on the stash for a large index

kevgrig:


mattzollinhofer:
This looks like a fairly straightforward change. Could you help me understand the use case for this? When does this change have the biggest impact?

mattzollinhofer:
Seems like this PR was an attempt to address #595 that didn’t quite work out. Do you think we can close this PR?

kevgrig:
@mattzollinhofer I just updated to the latest chewy Gem and I was able to run the `reset!` without an issue, thanks!

PR #598, opened 2017-10-31
Adding pipeline support to default_import_options

domitea:
Adding pipeline support to default_import_options for easier pipeline use in Type definition of Chewy Index.


pyromaniac:
👍  Thanks!

Issue #599, opened 2017-10-31
Newrelic logging returning undefined method `notice_sql` after upgrading newrelic_rpm to 4.x


Tongboy:
Hi,

newrelic logging as described in the readme (https://github.com/toptal/chewy#newrelic-integration) was working great on 3.17 - upgraded to 4.5.x and now I"m receiving the following stack

```
Failure/Error: SearchIndex.reset!

NoMethodError:
  undefined method `notice_sql' for #<NewRelic::Agent::TransactionSampler:0x0000000a1d3530>
  Did you mean?  notice_push_frame
# ./config/initializers/chewy.rb:47:in `block (2 levels) in <top (required)>'
# /home/rof/cache/bundler/ruby/2.3.0/gems/newrelic_rpm-4.5.0.337/lib/new_relic/agent/method_tracer.rb:74:in `block in trace_execution_scoped'
# /home/rof/cache/bundler/ruby/2.3.0/gems/newrelic_rpm-4.5.0.337/lib/new_relic/agent/method_tracer_helpers.rb:26:in `trace_execution_scoped'
# /home/rof/cache/bundler/ruby/2.3.0/gems/newrelic_rpm-4.5.0.337/lib/new_relic/agent/method_tracer.rb:72:in `trace_execution_scoped'
# ./config/initializers/chewy.rb:46:in `block in <top (required)>'
```

I reviewed the newrelic changelog and didn't notice any changes to the methods. anyone else having a similar problem or have any ideas?

Thank you!

barthez:
Hello @Tongboy,

Thanks for bringing it up!

I was updating application to New Relic 4.x and had the same trouble. I've dug into NewRelic RPM code and mimicked the way they do instrumentation for ActiveRecord, here is a code:

```ruby
require 'new_relic/agent/instrumentation/evented_subscriber'

class ChewySubscriber < NewRelic::Agent::Instrumentation::EventedSubscriber
  def start(name, id, payload)
    event = ChewyEvent.new(name, Time.current, nil, id, payload)
    push_event(event)
  end

  def finish(_name, id, _payload)
    pop_event(id).finish
  end

  class ChewyEvent < NewRelic::Agent::Instrumentation::Event
    OPERATIONS = {
      'import_objects.chewy' => 'import',
      'search_query.chewy' => 'search',
      'delete_query.chewy' => 'delete'
    }.freeze

    def initialize(*args)
      super
      @segment = start_segment
    end

    def start_segment
[...]

PR #601, opened 2017-11-03
Update Readme to fix broken newrelic integration

Tongboy:
Fixes newrelic_rpm status monitoring changes introduces in newrelic_rpm 4.x when upgrading from 3.x

related issue: #599 

all credit goes to barthez


pyromaniac:
Thanks!

PR #612, opened 2017-11-22
Prevent mapping explosions in specification field

mpospelov:
change specification field type to string in order to prevent issues.
Related issue #595


pyromaniac:
It actually was `string` and had been changed to object because `string` is more limited. And deprecated as well.

PR #620, opened 2017-12-21
Avoid Elasticsearch index updates when passing empty reflections

robertasg:
Passing `[]` or another equivalent of an empty reflection to `Index.import!` makes two requests to the corresponding Elasticsearch index. `GET /index` and `PUT /index` to ensure whether the index exists, and to update the index.

Given a complicated record which triggers many Chewy index updates on associations, this might result in N+1 requests, which might be undesired.

This change adds avoids these requests whenever the reflection is empty.

### DIFF
--- a/lib/chewy/type/import.rb
+++ b/lib/chewy/type/import.rb
@@ -126,6 +126,7 @@ module Chewy
       private
 
         def import_routine(*args)
+          return if args.first.blank? && !args.first.nil?
           routine = Routine.new(self, args.extract_options!)
           routine.create_indexes!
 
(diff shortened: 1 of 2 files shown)


lib/chewy/type/import.rb
```diff
@@ -126,6 +126,7 @@ def compose(object, crutches = nil, fields: [])
       private
 
         def import_routine(*args)
+          return if args.first.blank? && !args.first.nil?
```
pyromaniac: What if only the options are passed?
It probably makes sense to do this check after `args.extract_options!`

robertasg: In that case, given that options aren't empty, this would work the same as `Index.import!` (without objects, with imports everything under the scope). Right?

robertasg: I don't mind changing this to make it less confusing though.

PR #622, opened 2017-12-29
(WIP) Adapt to Elatic Search v6 changes

Chipairon:
- 'string' field type does not exist anymore. Changing
default_field_type to 'keyword'.
- Field property 'index' now receives a boolean. Keyword type is not
analyzed by default, so no need to specify explicitly.

@pyromaniac This still is not finished. The Journal index creation errors out about multi-index:
```
Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":{"root_cause":[{"type":"remote_transport_exception","reason":"[instance-some_id][some_ip][indices:admin/create]"}],"type":"illegal_argument_exception","reason":"Rejecting mapping update to [dev_chewy_stash] as the final mapping would have more than 1 type: [journal, specification]"},"status":400}
```
I am stuck there.

Also related, when resetting an index, even if I configure `journal: false` on settings, Chewy is trying to create the Journal index. I think it should not create it, as it is not going to be used if `journal: true`. Should be created on first use if not exists.

### DIFF
--- a/lib/chewy/config.rb
+++ b/lib/chewy/config.rb
@@ -82,7 +82,7 @@ module Chewy
       @disable_refresh_async = false
       @indices_path = 'app/chewy'
       @default_root_options = {}
-      @default_field_type = 'string'.freeze
+      @default_field_type = 'keyword'.freeze
       self.search_class = Chewy::Search::Request
     end
 
--- a/lib/chewy/stash.rb
+++ b/lib/chewy/stash.rb
@@ -10,17 +10,17 @@ module Chewy
     define_type :specification do
       default_import_options journal: false
 
-      field :value, index: 'no'
+      field :value
       field :specification, type: 'object', enabled: false
     end
 
     define_type :journal do # rubocop:disable Metrics/BlockLength
       default_import_options journal: false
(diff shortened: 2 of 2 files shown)


jduff:
I came across a few things when looking into this:

- You are correct that the journal index probably shouldn't be created when `journal: false`, but the bigger problem is that both the `journal` and `specification` mappings are in the same index, ES6 does not allow multiple mappings to be created in a single index anymore. I think this means that an index needs to be created for each of these.
- The next thing you will run into is that the `reset!` action [here](https://github.com/toptal/chewy/blob/master/lib/chewy/index/actions.rb#L192) will always try to apply the journal. I added `Chewy.configuration[:journal]` to the conditional and the task completed successfully. There are two possible fixes I see here: the conditional as I did or always create the journal index.

After those changes the deploy task succeeded.

I'm going to take a stab at fixing this in a PR, but I'm a ES novice so some feedback on my assessment and/or help with a fix would be awesome.

pyromaniac:
Solved in https://github.com/toptal/chewy/pull/630, check it out

PR #625, opened 2018-01-12
Rails::VERSION is more accurate check for presence of Rails

nattfodd:
`Rails` module can be defined without actual Rails gem.

I work on a Sinatra project, which has [ActionView](https://github.com/rails/rails/tree/master/actionview) in dependencies. `ActionView` itself does not depend on Rails, but it has [rails-html-sanitizer](https://github.com/rails/rails-html-sanitizer/blob/master/lib/rails/html/sanitizer.rb#L1) and [rails-dom-testing](https://github.com/rails/rails-dom-testing/blob/master/lib/rails/dom/testing/assertions.rb#L4) gems in dependencies, which both define `Rails` module.

Therefore, check `defined?(Rails)` isn't fully valid, and requires hacking on the client app side to make it work (i.e. defining `Rails.root` method and having `config/chewy.yml` file)

### DIFF
--- a/lib/chewy/config.rb
+++ b/lib/chewy/config.rb
@@ -155,7 +155,7 @@ module Chewy
 
     def yaml_settings
       @yaml_settings ||= begin
-        if defined?(Rails)
+        if defined?(Rails::VERSION)
           file = Rails.root.join('config', 'chewy.yml')
 
           if File.exist?(file)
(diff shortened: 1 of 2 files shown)


nattfodd:
Can someone check if those failing tests on CI are PR related, or those are floating specs? I ran those tests locally and they pass...

nattfodd:
Ok, I've investigated failing test case - it's definitely not related to the fix itself.

pyromaniac:
Thanks for the contribution!

PR #626, opened 2018-01-17
Do not underscore suggested index name

dm1try:
to allow legacy names without modifications

actual problem:
we wanna using`chewy` DSL just for querying the data at first
and our legacy name contains some dash symbols


pyromaniac:
Make sense, thanks!

PR #630, opened 2018-02-04
Split stash onto two indices and make it support ES6

pyromaniac:
In order to support ES6 we have to split an index with 2 types onto two singe-types indices.
Also, `string` type is not supported, so we use `keyword` with ES2 fallback.
Also, we want to store an unlimited amount of data in specifications because there are guys facing the limits. String data types are limited to several kb of data or so, object fields are limited to 1000 keys by default on a single level. So binary field is the only way to go.

### DIFF
--- a/lib/chewy/fields/base.rb
+++ b/lib/chewy/fields/base.rb
@@ -33,6 +33,13 @@ module Chewy
           end
         mapping.reverse_merge!(options)
         mapping.reverse_merge!(type: (children.present? ? 'object' : Chewy.default_field_type))
+
+        # This is done to support ES2 journaling and will be removed soon
+        if mapping[:type] == 'keyword' && Chewy::Runtime.version < '5.0'
+          mapping[:type] = 'string'
+          mapping[:index] = 'not_analyzed'
+        end
+
         {name => mapping}
       end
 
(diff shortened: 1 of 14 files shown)


lib/chewy/fields/base.rb
```diff
@@ -33,6 +33,13 @@ def mappings_hash
           end
         mapping.reverse_merge!(options)
         mapping.reverse_merge!(type: (children.present? ? 'object' : Chewy.default_field_type))
+
+        # This is done to ssuport ES2 journalling and will be removed soon
```
sergey-kintsel: support

lib/chewy/index/actions.rb
```diff
@@ -140,8 +135,15 @@ def purge!(suffix = nil)
...
         %i[import import!].each do |method|
           class_eval <<-METHOD, __FILE__, __LINE__ + 1
-            def #{method} options = {}
-              objects = options.reject { |k, v| !type_names.map(&:to_sym).include?(k) }
+            def #{method}(*args)
+              options = args.extract_options!
+              if args.one? && type_names.one?
+                objects = {type_names.first.to_sym => args.first}
+              elsif args.one?
+                fail ArgumentError, "Plase pass objects for `#{method}` as a hash with type names"
```
DNNX: Please

Issue #638, opened 2018-03-15
Ability to specify index name(s) for partitioning


jeremyhaile:
We are trying to do date based index partitioning (e.g. log-2018-01, log-2018-02)

When doing time based searches, we could then specify to only search the indexes that we know match the time range given but I can't find any way to specify this with chewy. 

I'm wanting to do something like `query = LogIndex.index('log-2018-01,log2018-02').filter(...)`

Is this supported now, or if not what is the best way to accomplish this?

pyromaniac:
Well, this is actually the first feature we need for ES6 and I'm thinking about it for a long time already but, unfortunately, it is not implemented yet. However, everything is actually ready for it in the DSL, the implementation will be simple if you want to try.

jeremyhaile:
@pyromaniac Happy to take a stab at it. Since I'm just figuring out the code base, is there any files in the code base you can recommend I look at that might be similar to what I will need to add? Appreciate it!

pyromaniac:
https://github.com/toptal/chewy/blob/master/lib/chewy/search/request.rb

Able1991:
+1

rabotyaga:
Handled by https://github.com/toptal/chewy/pull/657

PR #645, opened 2018-06-20
Support changing indices for query

cyu:
Changes for #638


rabotyaga:
Handled in https://github.com/toptal/chewy/pull/657

Issue #642, opened 2018-04-23
`.search_type` and `.preferences` always error


paulsuda:
I have a query like this, which works just fine...

```
UsersIndex.query(match: {name: 'test'}).object_hash
```

Following [the docs](https://github.com/toptal/chewy/blob/master/lib/chewy/search/request.rb#L403),  I tried something like this to use the user ID to guarantee the same shards respond for a give user...

```
UsersIndex.query(match: {name: 'test'}).preference('MY_USER_ID_STRING').object_hash
```
... this throws this error...

```
Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":{"root_cause":[{"type":"parsing_exception","reason":"Unknown key for a VALUE_STRING in [preference].","line":1,"col":15}],"type":"parsing_exception","reason":"Unknown key for a VALUE_STRING in [preference].","line":1,"col":15},"status":400}
from /home/vagrant/.rbenv/versions/2.3.6/lib/ruby/gems/2.3.0/gems/elasticsearch-transport-5.0.4/lib/elasticsearch/transport/transport/base.rb:202:in `__raise_transport_error'
```
... which is not that surprising, it looks like it's putting it in the body. This can be seen inspecting with `binding.pry`...

```
[42] pry(main)> UsersIndex.query(match: {name: 'test'}).preference('user_id_string')
=> <UsersIndex::Query {:index=>["users"], :type=>["user"], :body=>{:preference=>"user_id_string", :query=>{:match=>{:name=>"test"}}}}>
```

... same thing happens when I try to set a `search_type` like this according to [the docs](https://github.com/toptal/chewy/blob/master/lib/chewy/search/request.rb#L402)...

```
UsersIndex.query(match: {name: 'test'}).search_type(:dfs_query_then_fetch).object_hash
```
... it gives a similar error ...
```
Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":{"root_cause":[{"type":"parsing_exception","reason":"Unknown key for a VALUE_STRING in [search_type].","line":1,"col":16}],"type":"parsing_exception","reason":"Unknown key for a VALUE_STRING in [search_type].","line":1,"col":16},"status":400}
from /home/vagrant/.rbenv/versions/2.3.6/lib/ruby/gems/2.3.0/gems/elasticsearch-transport-5.0.4/lib/elasticsearch/transport/transport/base.rb:202:in `__raise_transport_error'
```
... also because it's putting `search_type` in the query body.

[...]

PR #647, opened 2018-08-25
Add query string parameter support

### LINKED ISSUE #586: request_cache calls do not work
Calling `UsersIndex.request_cache(true).render` returns:

```
{index=>['users'], :type=>['user'], :body=>{:request_cache=>true}}
```

However, the [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/shard-request-cache.html#_enabling_and_disabling_caching_per_request) say that the `request_cache` parameter needs to be passed as a "query-string parameter", not as a parameter in the body. Indeed, calling `UsersIndex.request_cache(true).total` returns:

```
Elasticsearch::Transport::Transport::Errors::BadRequest: [400] {"error":{"root_cause":
[{"type":"parsing_exception","reason":"Unknown key for a VALUE_BOOLEAN in [request_cache].",
"line:":1,"col":18}],"type":"parsing_exception","reason":"Unknown key for a VALUE_BOOLEAN in 
[request_cache].","line":1,"col":18,"status":400}
```

mattzollinhofer:
Per ES docs: search_type, request_cache, and the allow_partial_search_results
should not be in the body. They should be sent as query string parameters. (https://www.elastic.co/guide/en/elasticsearch/reference/6.4/search-request-body.html#_parameters_4)

My understanding is that chewy should act like this:
``` 
  $ IssuesIndex.request_cache(true).render
  # good (new)
  $ {:index=>["issues"], :type=>["issue"], :body=>{}, :request_cache=>true}
  # bad (old)
  $ {:index=>["issues"], :type=>["issue"], :body=>{:request_cache=>true}
```

I believe this resolves #586 and part of #642.

This is my first PR to this repo so please let me know if you'd like any changes. The second commit (Adding AllowPartialSearchResults parameter) is definitely optional as it supports an ES 6 feature. What do you think?

### DIFF
--- a/lib/chewy/search/parameters.rb
+++ b/lib/chewy/search/parameters.rb
@@ -10,7 +10,7 @@ module Chewy
     # @see Chewy::Search::Request#parameters
     # @see Chewy::Search::Parameters::Storage
     class Parameters
-      QUERY_STRING_PARAMS = %i[search_type request_cache].freeze
+      QUERY_STRING_PARAMS = %i[search_type request_cache allow_partial_search_results].freeze
       # Default storage classes warehouse. It is probably possible to
       # add your own classes here if necessary, but I'm not sure it will work.
       #
--- a/lib/chewy/search/parameters.rb
+++ b/lib/chewy/search/parameters.rb
@@ -10,6 +10,7 @@ module Chewy
     # @see Chewy::Search::Request#parameters
     # @see Chewy::Search::Parameters::Storage
     class Parameters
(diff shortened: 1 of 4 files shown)


pyromaniac:
Oh, I don't really understand this. https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-api supposed to handle this. But okay, let's merge it.

pyromaniac:
Thanks for the contribution!

PR #653, opened 2018-10-19
Extract type-related functionality to separate module

barthez:
With ES6 typeless indexes we want make index types optional. All type-related functionality in included to index only when it defines at least 1 type.


lib/chewy/query.rb
```diff
@@ -23,7 +23,7 @@ class Query
       timeout limit offset highlight min_score rescore facets script_score
       boost_factor weight random_score field_value_factor decay aggregations
       suggest none strategy query filter post_filter boost_mode
-      score_mode order reorder only types delete_all find total
+      score_mode order reorder only delete_all find total
```
pyromaniac: Why did you remove this?

lib/chewy/search/request.rb
```diff
@@ -26,7 +26,7 @@ class Request
         timeout min_score source stored_fields search_after
         load script_fields suggest aggs aggregations none
         indices_boost rescore highlight total total_count
-        total_entries types delete_all count exists? exist? find pluck
+        total_entries delete_all count exists? exist? find pluck
```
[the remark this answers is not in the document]
barthez: Method `types` was never delegated. Either `Chewy::Index` or `Chewy::Type` had `types` method defined. Right now, `types` method is present in Index class only if there is at least one type defined. With `types` being on the list, method is being delegated to search class via `method_missing`, which is not what we want. Also, when Index class is type-less we undefine `types` method from search class.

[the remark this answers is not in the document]
pyromaniac: This is the search request chainable method it filters types in the search query

lib/chewy.rb
```diff
@@ -113,6 +113,7 @@ def derive_type(name)
     # @raise [Chewy::UnderivableType] in cases when it is impossible to find index or type
     # @return [Array<Chewy::Type>] an array of derived types
     def derive_types(from)
+      return [] if from.is_a?(Class) && from < Chewy::Index && !from.has_types?
```
pyromaniac: This method should return index if the index is typeless.

pyromaniac: `create_type` method below needs to be moved to the new module.

bbatsov:
@barthez Is this PR still relevant?

rabotyaga:
> @barthez Is this PR still relevant?

@bbatsov No, I was keeping it for reference, just in case, as we are working on type removals. But it seems it can be safely closed.

PR #662, opened 2018-12-13
support jets application

byung82:


dalthon:
Hi @byung82, could you explain why this is needed?

dalthon:
I can't understand why this is needed and looks like it would break require of chewy on rails, so I am closing this unless we receive better explanations

PR #670, opened 2019-03-22
CI: Drop unused sudo:false setting

olleolleolle:
Removes an old setting from Travis. See https://blog.travis-ci.com/2018-11-19-required-linux-.infrastructure-migration


AlfonsoUceda:
We are not using travis CI anymore.

PR #689, opened 2019-09-30
Correctly report indices and aliases

mpeychich:
Improve the reporting on indices and aliases that match index naming.

### Before
Indices created via Chewy without a suffix would fail to report themselves as an index or alias for the Chewy index.
```ruby
DummyIndex.create! # => {'acknowledged' => true}
DummyIndex.indexes # => []
DummyIndex.aliases # => []
```

Indices created via Chewy with a suffix would correctly report the suffixed index but would fail to report the alias
```ruby
DummyIndex.create!('2013') # => {'acknowledged' => true}
DummyIndex.indexes # => ['dummies_2013']
DummyIndex.aliases # => []
```

### After
Indices created via Chewy without a suffix will correctly report themselves as an index.
```ruby
DummyIndex.create!
DummyIndex.indexes # => ['dummies']
DummyIndex.aliases # => []
```

Indices created via Chewy with a suffix will correctly report the alias
```ruby
DummyIndex.create!('2013') # => {'acknowledged' => true}
DummyIndex.indexes # => ['dummies_2013']
DummyIndex.aliases # => ['dummies']
```

### DIFF
--- a/lib/chewy/index/actions.rb
+++ b/lib/chewy/index/actions.rb
@@ -172,7 +172,7 @@ module Chewy
         def reset!(suffix = nil, apply_journal: true, journal: false, **import_options)
           result = if suffix.present?
             start_time = Time.now
-            indexes = self.indexes
+            indexes = self.indexes - [index_name]
             create! suffix, alias: false
 
             general_name = index_name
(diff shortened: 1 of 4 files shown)


dalthon:
It was rebased at https://github.com/toptal/chewy/pull/757 and some details were changed in order to avoid some false positives on indexes when two distinct classes are named with containing names.

It is waiting a little to merge, just to make sure that breaking changes introduced will not affect other stuff

PR #695, opened 2019-11-06
Clear the scroll by id after completing scroll_batches

amoslanka:
ElasticSearch has settings that limit the number of concurrent scrolls on the cluster. Scrolls are cleared automatically after the scroll ttl (eg, "1m") but I've found that iterating a lot of data in a lot of parallel jobs makes it easy to hit this limit. This may be avoided by staggering parallelism differently or by setting shorter scroll ttls, but it seems to also make sense to clear out the old scrolls when the use of scroll_batches is complete.


dalthon:
@amoslanka thank you for your PR, it was just rebased and merged!

PR #703, opened 2020-01-12
Update README.md: chewy:sync

orion122:


README.md
```diff
@@ -1013,7 +1013,7 @@ rake chewy:update[-users,places#city] # updates every index in the application e
 
 #### `chewy:sync`
 
-Provides a way to synchronize outdated indexes with the source quickly and without doing a full reset.
+Provides a way to synchronize outdated indexes with the source quickly and without doing a full reset (the field `updated_at` is required).
```
jonatas: This is a default column being generated on Rails. The framework also uses the defaults. Here we have a high-level overview of the sync rake task and not the specific requirements to make it work.

Maybe we can improve the source documentation here: https://github.com/toptal/chewy/blob/master/lib/chewy/type/mapping.rb#L12

orion122: I encountered the behavior of the framework when it does not use this field by default. I described the index without explicitly specifying updated_at:
```ruby
class UsersIndex < Chewy :: Index
   define_type User do
     field :first_name
     field :last_name
     field :created_at, type: 'date'
   end
end
```
Then I imported index - UsersIndex.import. This field was not imported to elasticsearch; accordingly, the synchronization feature did not work.

Then I explicitly indicated this field in the definition of the index, updated the index, and after that the synchronization worked.

dalthon: Actually it does not require necessarily `updated_at`, it requires field defined at method `outdated_sync_field` as described at source here: https://github.com/toptal/chewy/blob/d56adc43a0080f0f33330ae48215e76501a01db6/lib/chewy/type/syncer.rb#L3

By default it is `updated_at` if that field exists.

Issue #708, opened 2020-04-15
`UsersIndex.create!` not working with ES version 7.x


sachin1mittal:
I have a users_index class as following: 

`app/indexes/users_index.rb`

```
class UsersIndex < Chewy::Index
  define_type User.active do
    field :first_name
  end
end

```
With ES version 7.6.2, When I run `UsersIndex.create!`, it raises an error `"type": "mapper_parsing_exception",
        "reason": "Root mapping definition has unsupported parameters:` 
While when I do the same with version 6.8.8, it works fine.

I read that the body argument structure has been changed in ES version 7.

So, with version 7, request format should be:

```
PUT /users
{
  "mappings": {
      "properties": {
        "first_name": {
          "type": "text"
        }
    }
  }
}
```

While for earlier versions, it should be

```
PUT /users
{
  "mappings": {
     "user": {
        "properties": {
          "first_name": {
            "type": "text"
          }
      }
    }
  }
}
```

And, its failing because of this change.

I think changing `chewy-5.1.0/lib/chewy/fields/root.rb#33` from `mapping` to `mapping[name]` should make it work fine.

Basically, it's returning the mapping hash as 

```
{
    "user": {
        "properties": {
            "first_name": {
                "type": "text"
            }
        }
    }
}
```
While, for the 7.x versions, it should return as

```
 {
    "properties": {
        "first_name": {
            "type": "text"
        }
    }
}
```

przbadu:
I am facing same issue, after searching for a while, ES 7.x has removed mapping types https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html

przbadu:
@sachin1mittal is correct, returning `mapping[name]` is working and indexing mapping correctly for me.

silva96:
related to #673

reneklacan:
@sachalifs before it's fixed, you can try this nasty monkeypatch which fixes it for us on ES 7.8: https://gist.github.com/reneklacan/5ed4bd08b0ad43cd81cab38530ae0889

AlexWayfer:
> @sachalifs before it's fixed, you can try this nasty monkeypatch which fixes it for us on ES 7.8: https://gist.github.com/reneklacan/5ed4bd08b0ad43cd81cab38530ae0889

Maybe you want to create a PR with these fixes?

rabotyaga:
[...]

Issue #673, opened 2019-05-03
Elasticsearch 7.0 Removal of mapping types 


paul:
Re: https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html

I'm unable to use chewy to create indexes in Elasticsearch 7.0. Chewy attempts to create indexes with mappings that looks like this:

```
PUT /posts
{
  "settings": { "analysis": { ... } },
  "mappings" : {
    "post": { 
      "properties": { "id": { "type": "keyword" } , etc }
    }
  }
}
```

ES7 has removed the "mapping type" level from the "mappings" field, so now expects documents that look like this:

```
{
  "settings": { "analysis": { ... } },
  "mappings" : {
    "properties": { "id": { "type": "keyword" } , etc }
  }
}
```

For the 7.x series, the old way is still supported, but only my appending a `?include_type_name=true` query param to the create index URL.

```
PUT posts?include_type_name=true
```

I tried a few ways to set it in the body, but that doesn't seem possible. It also doesn't seem possible for chewy to provide a query param when creating the index.

noellabo:
The following is an example of a workaround with compatibility parameters.
https://github.com/noellabo/chewy/commit/94e8a6cfc540e06c051d1217e13dda0d73442e8b

As it is possible to respond by rewriting the URL, some people have adopted a method to solve using nginx.
https://gist.github.com/Vincanote/2b18ba8aa33d068a9cc012b9a4d891af

Dee-2015:
Just linking to my other comment 

https://github.com/tootsuite/mastodon/issues/10664#issuecomment-507057197

StephRicardo:
Is there a plan to update Chewy to work with ES 7 or to use the workaround for now?

longnd:
hi @pyromaniac, could you pls give a clarification on how to use chewy with ES 7 ?

pyromaniac:
Hello everyone. There is definitely a plan to update Chewy. Unfortunately, I can't tell when it is going to happen exactly. Hopefully, sooner than later. I'm sorry for it.

silva96:
Hi, is there at least a branch where we can coordinate this? I would like to contribute.

darkleaf:
```ruby
# config/initializers/elastic_7.rb
# https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#hits-total-now-object-search-response
[...]

PR #712, opened 2020-04-22
Add ES 7 support

przbadu:
#708 

support for Elastic Search 7.x mappings

### DIFF
--- a/lib/chewy/fields/root.rb
+++ b/lib/chewy/fields/root.rb
@@ -30,7 +30,7 @@ module Chewy
         end
 
         mappings[name][:_parent] = parent.is_a?(Hash) ? parent : {type: parent} if parent
-        mappings
+        mappings[name]
       end
 
       def dynamic_template(*args)


silva96:
will this break the compatibility with the older versions? it might need a condition to verify how does the response format look like doesn't it?

subelsky:
I'm testing Chewy locally and am successfully using `przbadu`'s code w/ an elasticsearch 7.0 FWIW. 👍  this was a lifesaver!

lib/chewy/fields/root.rb
```diff
@@ -30,7 +30,7 @@ def mappings_hash
         end
 
         mappings[name][:_parent] = parent.is_a?(Hash) ? parent : {type: parent} if parent
-        mappings
+        mappings[name]
```
darkleaf, line 33: It is the breaking change. 
If you want to support both 6 and 7 Elastic versions you could pass special request params.

 https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#hits-total-now-object-search-response
 https://www.elastic.co/blog/moving-from-types-to-typeless-apis-in-elasticsearch-7-0
 https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html#_schedule_for_removal_of_mapping_types

https://github.com/toptal/chewy/issues/673#issuecomment-631432562

przbadu (commented):
But I think, because it is a breaking change in ElasticSearch version, maybe we can have newer version of chewy handling ES 7+ and we can have previous version for old ES version?

TrevorHinesley:
Unfortunately, this doesn't allow querying on types in an index. 

https://github.com/toptal/chewy#composing-requests

^ the example of `PlaceIndex::City` below will return no results:

```ruby
PlaceIndex.query(match: {name: 'London'}) # returns documents of any type
PlaceIndex::City.query(match: {name: 'London'}) # returns cities only.
```

bbatsov:
I'm willing to accept this patch, but only if it doesn't break compatibility with older ES versions.

lib/chewy/fields/root.rb
silva96: How does the response look like for pre-7 and for 7+ ? based on that, you could determine if you need to return either `mapping` or `mapping[name]` 

or am I oversimplifying the problem and I'm not taking something in account?

rabotyaga:
Please take a look at [Chewy 7.0.0 release](https://github.com/toptal/chewy/releases/tag/v7.0.0). Stay tuned for the 7.1 release, with dropped support for ES6.8 and removed deprecation warnings.

PR #720, opened 2020-08-07
Symbolize sidekiq options

abraham-chan:
Sidekiq will convert symbols to strings when serializing arguments.
(See https://github.com/mperham/sidekiq/issues/3893 and https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple)

However, classes such as Routine use keyword arguments as initializers and will throw an argument error if provided options with string keys. (See https://github.com/toptal/chewy/blob/v5.0.0/lib/chewy/type/import/routine.rb#L44)

For example:
```ruby
# With symbol options
CitiesIndex::City.import!([1], update_fields: [:name])
=> true

# With string options
CitiesIndex::City.import!([1], "update_fields" => [:name]) 
ArgumentError: wrong number of arguments (given 2, expected 1)
```

A simple solution is to symbolize the options before running `import!`

### DIFF
--- a/lib/chewy/strategy/sidekiq.rb
+++ b/lib/chewy/strategy/sidekiq.rb
@@ -15,7 +15,7 @@ module Chewy
 
         def perform(type, ids, options = {})
           options[:refresh] = !Chewy.disable_refresh_async if Chewy.disable_refresh_async
-          type.constantize.import!(ids, options)
+          type.constantize.import!(ids, options.symbolize_keys)
         end
       end
 
--- a/spec/chewy/strategy/sidekiq_spec.rb
+++ b/spec/chewy/strategy/sidekiq_spec.rb
@@ -39,6 +39,18 @@ if defined?(::Sidekiq)
     end
 
     specify do
+      expect(CitiesIndex::City).to receive(:import!).with([city.id.to_s], suffix: '201601')
+      ::Sidekiq::Testing.fake! do
+        Chewy::Strategy::Sidekiq::Worker.perform_async(
+          'CitiesIndex::City',
+          [city.id.to_s],
+          'suffix' => '201601'
+        )
+      end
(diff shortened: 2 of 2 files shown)


rabotyaga:
Thanks @abraham-chan !
Is there any legit way of getting keyword options in the `Chewy::Strategy::Sidekiq::Worker` `args` list?

abraham-chan:
> Thanks @abraham-chan !
> Is there any legit way of getting keyword options in the `Chewy::Strategy::Sidekiq::Worker` `args` list?

I don't believe so. [Their documentation](https://github.com/mperham/sidekiq/wiki/Best-Practices#1-make-your-job-parameters-small-and-simple) is pretty clear
> This means you must not use ruby symbols as arguments. 
> Don't pass symbols, named parameters, keyword arguments or complex Ruby objects (like Date or Time!) as those will not survive the dump/load round trip correctly.

rabotyaga:
I was trying to ask something different: how you can end up with keyword options passed to the `Chewy::Strategy::Sidekiq::Worker` `args` list? Or, in other words, what real-life scenario does this PR address?

abraham-chan:
> I was trying to ask something different: how you can end up with keyword options passed to the `Chewy::Strategy::Sidekiq::Worker` `args` list? Or, in other words, what real-life scenario does this PR address?

Oh I see, my mistake. We were directly calling `Chewy::Strategy::Sidekiq::Worker.perform_async` in a few places in our code, when we wanted an async import for specific fields. We've been using the `symbolize_keys` fix in production for some time now with no issues.

e.g.
```
ChewyWorkers::ChewyWorker.perform_async("ProductsIndex", [1,2,3], update_fields: ["name"])
```

rabotyaga:
Please upgrade to chewy 5.2.0, the issue is solved in that version.

PR #734, opened 2021-01-27
Add support for Ruby 3

lowang:
#### Testing:
- switch to ruby 3
- `rspec`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,8 +22,7 @@ workflows:
       - rspec-ruby-27-activerecord60
       - rspec-ruby-27-activerecord52
       - rspec-ruby-27-mongoid
-        # TODO elastic 7
-        # TODO sequel?
+      - rspec-ruby-30-activerecord61
 
 commands:
   rspec-test:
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -23,6 +23,7 @@ workflows:
       - rspec-ruby-27-activerecord60
       - rspec-ruby-27-activerecord52
       - rspec-ruby-27-mongoid
+      - rspec-ruby-30-activerecord60
         # TODO elastic 7
         # TODO sequel?
```
AlfonsoUceda, line 28: I would remove these TODOs for now.

AlfonsoUceda (approved):

AlexVPopov (approved):

README.md
```diff
@@ -92,7 +92,9 @@ Or install it yourself as:
 
 ### Ruby
 
-Chewy is compatible with MRI 2.4-2.7. Ruby 3 support is currently in development.
+Chewy is compatible with MRI 2.4-3.0. [^1]
```
barthez, line 95: btw. 2.4 ended its life last year, we could get rid of it.

barthez (approved):

rabotyaga (approved):

alishersadikov:
> > Out of curiosity, why is Ruby 3 noted as only being supported on Rails 6.1?
> 
> actually it should be `>= 6.1`, and this is only because rails prior to 6.1 doesn't support ruby >= 3.0.0

I am not sure this is accurate any more - maybe it was at the time of this post? Rails 6.0.1 seems to support Ruby 3: 
https://rubygems.org/gems/rails/versions/6.0.1

Also here: 
https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html

Not sure if you meant the `recommended` Ruby version is 3 on Rails 6.1.

PR #736, opened 2021-01-28
Fix nil children when using witchcraft

taylor-au:
The functionality of Chewy without witchcraft enabled simply nulls out any children fields where the child is null. With witchcraft enabled, Chewy will try to traverse objects even when a child is null. Code change addresses this issue.


mrzasa (approved):
:+1:

rabotyaga (approved):

rabotyaga:
Thanks, @taylor-au !

taylor-au:
@rabotyaga No problem, thanks for the great library!

PR #735, opened 2021-01-27
Patch ElasticSearch deprecation warning

Bhacaz:
Hi, since some peoples seems to work/wish to make Chewy compatible with ElasticSearch 6-7, here is a small PR to fix the deprecation warning from ElasticSearch. It seems to be a good first step.

PR #740, opened 2021-01-29
Prevent calling Index#delete with an empty string

Bhacaz:
Bug introduced with #735

ES client return an `Elasticsearch::Transport::Transport::Errors::BadRequest` error, which is not rescued. To keep the same behaviour, I simply call the ES client with the `index_name`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Bug fixes
 
+  * [#740](https://github.com/toptal/chewy/pull/740): Prevent calling Index#delete when the index does not exist. ([@bhacaz][])
   * [#736](https://github.com/toptal/chewy/pull/736): Fix nil children when using witchcraft ([@taylor-au][])
 
 ## Changes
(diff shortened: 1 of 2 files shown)


rabotyaga:
@Bhacaz Thank you very much!
Could you please describe how to reproduce this bug in a little bit more detailed way?
Or add a spec, that would fail without changes in `lib/chewy/index/actions.rb`. Added in this PR spec successfully passes on master.

Bhacaz:
I didn’t dig enough about the error. An `Elasticsearch::Transport::Transport::Errors::BadRequest` will only occurs in ES 1.X (which is the version I currently working to upgrade). Here what I found just to keep it this PR well documented.

If the index doesn't exists (or already destroyed)

### ES 1.X

1. `get_alias` return `{}`
2. `{}.keys.join('') # => ''`
3. `delete(index: '')` throw `Elasticsearch::Transport::Transport::Errors::BadRequest`
4. raise the error because it's not rescued

### ES 5.X and above

1. `get_alias` throw `Elasticsearch::Transport::Transport::Errors::NotFound`
2. `Elasticsearch::Transport::Transport::Errors::NotFound` is rescued

---

May I suggest 2 options:

1. Closing the PR, because it's a problem only in ES 1.X.
2. Adding a safe guard at the beginning of `Chewy::Index::Actions::ClassMethods#delete` to return false if the index doesn't existe.

rabotyaga:
@Bhacaz Thanks, let's close this for now.

PR #741, opened 2021-01-29
Document string vs keyword/text for ES6

mrzasa:
The only usage of `string` that I found in chewy was hidden in `Chewy::Runtime.version < '5.0'`. I've updated comments and started a file with migration guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Or install it yourself as:
 
 ### Ruby
 
-Chewy is compatible with MRI 2.5-3.0¹. 
+Chewy is compatible with MRI 2.5-3.0¹.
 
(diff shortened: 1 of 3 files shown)


AlfonsoUceda (approved):

.circleci/config.yml
```diff
@@ -238,3 +238,27 @@ jobs:
       GEMFILE: gemfiles/rails.6.1.activerecord.gemfile
     steps:
       - rspec-test
+
+  rspec-ruby-27-activerecord61-es6:
```
rabotyaga, line 242: Seems this should be removed, as it duplicates names from the feature branch

rabotyaga (approved):

PR #749, opened 2021-02-04
Avoid importing everything when given an empty relation

dalthon:
Just a rebase of #698 with one more spec added to cover empty relations.


rabotyaga (approved):
Please add a CHANGELOG entry

PR #687, opened 2019-09-26
Add multi search feature

mpeychich:
The Elasticsearch [Multi Search API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-multi-search.html) can be used to issue multiple search queries at a time.

PR #751, opened 2021-02-08
Add Chewy::MultiSearch Feature

dalthon:
Rebase of https://github.com/toptal/chewy/pull/687

### DIFF
--- a/spec/chewy/multi_search_spec.rb
+++ b/spec/chewy/multi_search_spec.rb
@@ -8,19 +8,19 @@ describe Chewy::MultiSearch do
     stub_model(:city)
     stub_model(:country)
 
-    stub_index(:places) do
+    stub_index(:cities) do
+      def self.aggregate_by_country
+        aggs(country: {terms: {field: :country_id}})
+      end
+
       define_type City do
         field :name, type: 'keyword'
         field :country_id, type: 'keyword'
-
-        def self.aggregate_by_country
-          aggs(country: {terms: {field: :country_id}})
-        end
       end
     end
   end
 
-  let(:places_query) { PlacesIndex.all }
+  let(:places_query) { CitiesIndex.all }
 
   describe '#queries' do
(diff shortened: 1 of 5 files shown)


spec/chewy/multi_search_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        multi_search.add_query(places_query)
+      end.to change {
+        multi_search.queries
+      }.from([]).to([places_query])
+    end
+  end
+
+  context 'when given two queries' do
+    let(:queries) { [aggregates, results] }
+    let(:aggregates) { PlacesIndex::City.aggregate_by_country.limit(0) }
```
AlfonsoUceda, line 45: Let's remove the type City, it will work and we won't have to change it for ES6

spec/chewy/multi_search_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+describe Chewy::MultiSearch do
+  before { Chewy.massacre }
+
+  before do
+    stub_model(:city)
+    stub_model(:country)
+
+    stub_index(:places) do
+      define_type City do
+        field :name, **KEYWORD_FIELD
```
AlfonsoUceda, line 13: Better replace this constant with
```suggestion
        field :name, type: 'keyword'
```

AlfonsoUceda (approved):

spec/chewy/multi_search_spec.rb
dalthon: Done!

spec/chewy/multi_search_spec.rb
dalthon: Ok! Done!

dalthon:
@rabotyaga and @AlfonsoUceda pointed out that it would be nice to not use types, so I will be better to wait until until we have 6.x suppoort.

PR #752, opened 2021-02-08
Remove multiple types mappings support

rabotyaga:
Remove support for the multiple types in mappings.

### DIFF
--- a/lib/chewy/index.rb
+++ b/lib/chewy/index.rb
@@ -146,6 +146,8 @@ module Chewy
       # passed.
       #
       def define_type(target, options = {}, &block)
+        raise 'Multiple types are deprecated' if type_hash.present?
+
         type_class = Chewy.create_type(self, target, options, &block)
         self.type_hash = type_hash.merge(type_class.type_name => type_class)
       end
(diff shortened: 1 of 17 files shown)


lib/chewy/index.rb
```diff
@@ -146,6 +146,8 @@ def prefix
       # passed.
       #
       def define_type(target, options = {}, &block)
+        raise 'Multiple types are deprecated' if type_hash.present?
```
AlfonsoUceda, line 149: Instead of raise if type_hash is present, we could raise when a user is trying to set more than one type.
```suggestion
        raise "Multiple types aren't supported by ElasticSearch v6" if type_hash.size > 1
```
I suggest this or it will fail in the projects.

AlfonsoUceda (approved):

lib/chewy/index.rb
rabotyaga: `if type_hash.present?` guard will be triggered on the 2nd invocation of `define_type` in some particular index class. While `if type_hash.size > 1` will allow to declare 2 types in 1 index.

PR #637, opened 2018-02-23
Add support for direct_import parameter to skip loading default scope on import

TikiTDO:
Proof of Concept to get around issue outlined in #636. 

Currently there's no tests, as I wasn't sure how the spec files were structured for this project. I can add the specs if someone could point me to an example where I could exercise the functionality of importing objects with extra includes.

PR #753, opened 2021-02-08
Add support for direct_import parameter to skip objects reloading

dalthon:
Just a rebase of https://github.com/toptal/chewy/pull/637 with added spec and a small fix regarding id types

### DIFF
--- a/lib/chewy/type/adapter/orm.rb
+++ b/lib/chewy/type/adapter/orm.rb
@@ -45,6 +45,7 @@ module Chewy
         # Import options:
         #
         #   <tt>:batch_size</tt> - import batch size, 1000 objects by default
+        #   <tt>:direct_import</tt> - import objects without reloading
         #
         # Method handles destroyed objects as well. In case of objects ORM scope
         # or array passed, objects, responding with true to `destroyed?` method will be deleted
@@ -75,10 +76,10 @@ module Chewy
         def import(*args, &block)
           collection, options = import_args(*args)
 
-          if collection.is_a?(relation_class)
-            import_scope(collection, options, &block)
-          else
+          if !collection.is_a?(relation_class) || options[:direct_import]
             import_objects(collection, options, &block)
+          else
+            import_scope(collection, options, &block)
           end
         end
 
@@ -119,6 +120,8 @@ module Chewy
           indexed = collection_ids.each_slice(options[:batch_size]).map do |ids|
             batch = if options[:raw_import]
               raw_default_scope_where_ids_in(ids, options[:raw_import])
+            elsif options[:direct_import]
+              hash.values_at(*ids.map(&:to_s))
             else
               default_scope_where_ids_in(ids)
             end
--- a/lib/chewy/type/import.rb
+++ b/lib/chewy/type/import.rb
@@ -67,6 +67,7 @@ module Chewy
         # @option options [String] suffix an index name suffix, used for zero-downtime reset mostly, no suffix by default
         # @option options [Integer] bulk_size bulk API chunk size in bytes; if passed, the request is performed several times for each chunk, empty by default
         # @option options [Integer] batch_size passed to the adapter import method, used to split imported objects in chunks, 1000 by default
+        # @option options [Boolean] direct_import skips object reloading in ORM adapter, `false` by default
         # @option options [true, false] journal enables imported objects journaling, false by default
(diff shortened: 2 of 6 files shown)


lib/chewy/type/adapter/orm.rb
```diff
@@ -45,6 +45,7 @@ def identify(collection)
         # Import options:
         #
         #   <tt>:batch_size</tt> - import batch size, 1000 objects by default
+        #   <tt>:direct_import</tt> - use the collectiov values as-is, instead of reloading
```
rabotyaga, line 48: The wording in https://github.com/sideqik/chewy/commit/21df05b77eb8ec72cf09811ea4dcd8919080c4f6 seems more clear to me

lib/chewy/type/import.rb
```diff
@@ -67,6 +67,7 @@ module ClassMethods
         # @option options [String] suffix an index name suffix, used for zero-downtime reset mostly, no suffix by default
         # @option options [Integer] bulk_size bulk API chunk size in bytes; if passed, the request is performed several times for each chunk, empty by default
         # @option options [Integer] batch_size passed to the adapter import method, used to split imported objects in chunks, 1000 by default
+        # @option options [Boolean] direct_import passed to the adapter import method, used skip the collection reload
```
rabotyaga, line 70: The wording in https://github.com/sideqik/chewy/commit/21df05b77eb8ec72cf09811ea4dcd8919080c4f6 seems more clear to me

spec/chewy/type/adapter/active_record_spec.rb
```diff
@@ -86,7 +86,19 @@ def import(*args)
...
-      specify { expect(import(cities)).to eq([{index: cities}]) }
+      specify do
+        cities
+        expects_db_queries do
+          expect(import(cities, direct_import: false)).to eq([{index: cities}])
+        end
+      end
+      specify do
+        cities
+        expects_no_query do
```
rabotyaga, line 97: Does it really fail if some db queries are made? Doesn't seem so.

rabotyaga (changes requested):

spec/chewy/type/adapter/active_record_spec.rb
dalthon: Changed! Now it does! While linting I removed the raise keyword by mistake

lib/chewy/type/import.rb
```diff
@@ -67,6 +67,7 @@ module ClassMethods
         # @option options [String] suffix an index name suffix, used for zero-downtime reset mostly, no suffix by default
         # @option options [Integer] bulk_size bulk API chunk size in bytes; if passed, the request is performed several times for each chunk, empty by default
         # @option options [Integer] batch_size passed to the adapter import method, used to split imported objects in chunks, 1000 by default
+        # @option options [Boolean] import objects without reloading
```
rabotyaga, line 70:
```suggestion
        # @option options [Boolean] direct_import skips object reloading in ORM adapter
```

rabotyaga: or
```suggestion
        # @option options [Boolean] direct_import import objects without reloading
```

rabotyaga (commented):
Should we add the option to the `Chewy::Type::IMPORT_OPTIONS_KEYS` ? I think otherwise it would be impossible to use it in `default_import_options`.

dalthon:
@rabotyaga I was checking values at `Chewy::Type::IMPORT_OPTIONS_KEYS` then I realized that those are not matching with method `import` options:

```diff
< consistency
> direct_import
< pipeline
< raw_import
< refresh
< replication
> parallel
> suffix
> update_failover
> update_fields
```

Do you think that the same applies for `parallel`, `suffix`, `update_failover` and `update_fields`?

rabotyaga:
@dalthon `refresh`, `consistency`, `replication` & `pipeline` belong to `Routine`: https://github.com/toptal/chewy/blob/7efd2e34b2a01cc2b4f72f9bb0a076825787cc36/lib/chewy/type/import/routine.rb#L25-L30
`raw_import` is passed to form adapter.
Let's leave missing `import` keys like `update_failover` & `update_fields` since nobody complains.

And let's add CHANGELOG entry.

dalthon:
@rabotyaga changelog added

PR #757, opened 2021-02-11
Correctly report indices and aliases

dalthon:
Just a rebased version of https://github.com/toptal/chewy/pull/689

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Changes
 
-  * [#757](https://github.com/toptal/chewy/pull/757): Correctly report indices and aliases ([@mpeychich][], [@dalthon][])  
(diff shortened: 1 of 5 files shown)


rabotyaga (changes requested):
What will happen, if we will have `UsersIndex` & `UsersActivitiesIndex`?

dalthon:
@rabotyaga as you guessed correctly it breaks... let me check if I can quickly solve this, otherwise I think it might be good to close this PR

dalthon:
@rabotyaga unfortunately we have to rely on something other than name prefix to associate Index class and indexes. This may be done by aliases or perfect name match.

Not every cases are covered, at https://github.com/toptal/chewy/blob/7edd7ae7656758d6cbd66122978680470e952dd5/spec/chewy/index/actions_spec.rb#L64 and https://github.com/toptal/chewy/blob/7edd7ae7656758d6cbd66122978680470e952dd5/spec/chewy/index/actions_spec.rb#L123 I highlighted cases where, due to the lack of alias, we can't map all indexes.

But I believe that is better to have false positives than false negatives in that unlikely case.

WDYT?

rabotyaga (approved):

barthez (approved):

PR #778, opened 2021-03-26
Ignore empty blank options on fields

Vitalina-Vakulchyk:
Add :ignore_blank option to not pass fields when we have field responds true to .empty? method

* [x] ~~Commit message starts with `[Fix #issue-number]` (if the related issue exists).~~
* [x] Updated tests.


lib/chewy/fields/base.rb
```diff
@@ -53,6 +55,18 @@ def compose(*objects)
 
     private
 
+      def field_with_ignore_blank_flag?(result)
+        @options[:ignore_blank] == true && field_with_empty_value?(result)
+      end
```
dalthon, line 60: This should work as well and is shorter

```suggestion
      def field_with_ignore_blank_flag?(result)
        @options[:ignore_blank] && result.blank?
      end
```

lib/chewy/fields/base.rb
```diff
@@ -53,6 +55,18 @@ def compose(*objects)
...
+
+      def geo_point_field_without_ignore_blank_flag?(result)
+        !@options[:ignore_blank] && @options[:type] == :geo_point && field_with_empty_value?(result)
+      end
```
dalthon, line 64: here we should take a look at 2 things:

#### First:

Let's assume that someone really wants raise an error when field with geo_point type is blank (because it was supposed to be always filled or any other reason). If they explicitly pass option `ignore_blank: false` we should expect him to receive an error if field is blank, don't you think?

This case also should be tested, and we assume that the expected behavior is to raise an error.

#### Second:

Instead of checking for `@options[:type] == :geo_point` we should check `@options[:type].to_sym == :geo_point`
to allow users to pass types as string.

lib/chewy/fields/base.rb
```diff
@@ -53,6 +55,18 @@ def compose(*objects)
...
+
+      def field_with_empty_value?(result)
+        !result.nil? && result.empty?
+      end
```
dalthon, line 68: You can remove this method, since [blank? method](https://guides.rubyonrails.org/active_support_core_extensions.html#blank-questionmark-and-present-questionmark) from active support will behave exactly as you are expecting

Gemfile
```diff
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 
 gemspec
 
+gem 'pry'
```
dalthon, line 5: I guess you forgot to remove this

spec/support/active_record.rb
```diff
@@ -19,9 +20,17 @@
   create_table :cities do |t|
     t.column :country_id, :integer
     t.column :name, :string
+    t.column :surname, :string
```
dalthon, line 23: I will propose a not so important stuff: what about calling this new column `mayor_name` or something else?

It felt a little odd to me a city having surname (I also would not be impressed if there are cities somewhere with surnames...)

spec/chewy/fields/base_spec.rb
```diff
@@ -402,33 +402,234 @@
...
+          end
+        end
+
+        # this way of sending {} as location throws 'ArgumentError: wrong number of arguments (given 0, expected 1..2)'
+        # for Ruby 3.0.0
+        # in ./lib/chewy/fields/root.rb:72
+        #
+        # Failure/Error:
+        # def compose(object, crutches = nil, fields: [])
+        # ...
```
dalthon, line 615: Ops... We forgot to remove that

spec/chewy/fields/base_spec.rb
```diff
@@ -402,33 +402,234 @@
...
+
+      xcontext 'geo_point type with ignore_blank: true flag' do
+        before do
+          stub_index(:countries) do
+            define_type Country do
+              field :id
+              field :cities do
+                field :id
+                field :name
+                field :location, type: :geo_point, ignore_blank: true do
```
dalthon, line 559: As I mentioned before, it would be nice to test this combination as well: `type: :geo_point, ignore_blank: false`

spec/chewy/fields/base_spec.rb
```diff
@@ -402,33 +402,234 @@
...
+              end
+            end
+          end
+        end
+
+        let(:country_with_cities) do
+          cities = []
+
+          Country.create!(id: 1, cities: cities)
+        end
```
dalthon, line 543: You meant `:country_without_cities`?

Otherwise, ActiveRecord empty has_many associations would return empty relation, you don't need to pass cities as an empty array

```suggestion
        let(:country_with_cities) { Country.create!(id: 1) }
```

spec/chewy/fields/base_spec.rb
```diff
@@ -402,33 +402,234 @@
...
+            .to eq('id' => 1)
+        end
+
+        specify do
+          expect(CountriesIndex::Country.root.compose(country_with_cities)).to eq(
+            'id' => 1, 'cities' => [
+              {'id' => 1, 'name' => '', 'description' => nil}
+            ]
+          )
+        end
```
dalthon, line 521: I would replace this with something like:

```suggestion
        let(:country) { Country.create!(id: 1, cities: cities) }

        context('without cities') do
          let(:cities) { [] }

          specify do
            expect(CountriesIndex::Country.root.compose(country))
              .to eq('id' => 1)
          end
        end

        context('with cities') do
          let(:cities) { [City.create!(id: 1, name: '', surname: '') ] }

          specify do
            expect(CountriesIndex::Country.root.compose(country)).to eq(
              'id' => 1, 'cities' => [
                {'id' => 1, 'name' => '', 'description' => nil}
              ]
            )
          end
        end
```

lib/chewy/fields/base.rb
Vitalina-Vakulchyk: Thank you!

About the 1st comment: 
I added the spec for 
`field :location, type: :geo_point, ignore_blank: false`.

And
{'id' => 1, 'name' => 'City1', 'location' => {}},
{'id' => 2, 'name' => 'City2', 'location' => ''}

returns

{'id' => 1, 'location' => {}, 'name' => 'City1'},
{'id' => 2, 'location' => '', 'name' => 'City2'}

and does not return an error

spec/support/active_record.rb
Vitalina-Vakulchyk: You're right :)

spec/chewy/fields/base_spec.rb
Vitalina-Vakulchyk: Thank you!

I have added this spec and described its output above.

Vitalina-Vakulchyk:
Thank you, Dalton! @dalthon 
Please take a look again.

dalthon (approved):
🚀 Great!

Just left an opinionated suggestion that you can easily ignore if you don't like it.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ### Changes
 
+  * [#778](https://github.com/toptal/chewy/pull/778): Add ignore_blank flag for field method ([@Vitalina-Vakulchyk][]):
```
AlexVPopov, line 9:
```suggestion
  * [#778](https://github.com/toptal/chewy/pull/778): Add `ignore_blank` option to `field` method ([@Vitalina-Vakulchyk][]):
```

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
...
+    * true by default for geo_point type
```
AlexVPopov, line 10:
```suggestion
    * `true` by default for `geo_point` type
```

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
...
+    * false by default for other types
```
AlexVPopov, line 11:
```suggestion
    * `false` by default for other types
```

README.md
```diff
@@ -660,6 +662,25 @@ By default, when you perform import Chewy checks whether an index exists and cre
 You can turn off this feature to decrease Elasticsearch hits count.
 To do so you need to set `skip_index_creation_on_import` parameter to `false` in your `config/chewy.yml`
 
+### Skip record fields during import
+
+You can use `ignore_blank: true` to skip fields, that return `true` for `.empty?` method:
```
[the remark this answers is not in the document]
AlexVPopov: I agree with Dalton's suggestion inside the parenthesis, but think that we could keep the geo point section as well, just to be explicit.

lib/chewy/fields/base.rb
```diff
@@ -40,6 +40,8 @@ def mappings_hash
       def compose(*objects)
         result = evaluate(objects)
 
+        return {} if field_with_ignore_blank_flag?(result) || geo_point_field_without_ignore_blank_flag?(result)
```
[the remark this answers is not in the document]
dalthon: The thing here is that it would be nice to have a distinct default value for geo points (because blank values can't be indexed and raises an error).

But it would be also nice that, in case someone really wants to not ignore blanks and expect to have errors raised, I believe we should offer them this option.

Also remember that `''.blank?` and `[].blank?` returns true and we may want to have normal fields with empty strings or arrays indexed with those blank values.

Also notice that at method `geo_point_field_without_ignore_blank_flag?` she is checking `@options[:ignore_blank] != false` instead of `!@options[:ignore_blank]`, having default as `nil`, explicitly checking for not `false` handles this default without any extra effort

lib/chewy/fields/base.rb
```diff
@@ -53,6 +55,14 @@ def compose(*objects)
...
     private
 
+      def field_with_ignore_blank_flag?(result)
+        @options[:ignore_blank] == true && result.blank?
+      end
+
+      def geo_point_field_without_ignore_blank_flag?(result)
+        @options[:ignore_blank] != false && @options[:type] && @options[:type].to_sym == :geo_point && result.blank?
+      end
+
```
barthez, line 65:
```suggestion
   def geo_point?
     @options[:type].to_s == 'geo_point'
   end
   
   def ignore_blank?
     @options.fetch(:ignore_blank) { geo_point? }
   end
```

then you could use:
```ruby
return {} if result.blank? && ignore_blank?
```

barthez: This would work only if you don't expect to ignore `ignore_blank` options for `geo_point` field.

spec/chewy/fields/base_spec.rb
```diff
@@ -402,33 +402,279 @@
...
+        Location.belongs_to :city
+        City.has_one :location, -> { order :id }
         Country.has_many :cities, -> { order :id }
+      end
 
-        stub_index(:countries) do
-          define_type Country do
-            field :id
-            field :cities do
+      context 'text fields type with and without ignore_blank flag' do
```
rabotyaga, line 415: Why this is inside the 'objects and scopes' context?

spec/chewy/fields/base_spec.rb
```diff
@@ -402,33 +402,279 @@
...
+              {'id' => 1, 'name' => '', 'historical_name' => ''},
+              {'id' => 2, 'name' => '', 'historical_name' => ''}
+            ]
+          )
+        end
       end
 
-      let(:country_with_cities) do
-        cities = [City.create!(id: 1, name: 'City1'), City.create!(id: 2, name: 'City2')]
+      context 'fields with ignore_blank: true flag with .empty? method' do
```
rabotyaga, line 449: `empty?` or `blank?`?

spec/chewy/fields/base_spec.rb
```diff
@@ -402,33 +402,279 @@
...
+            ]
+          )
+        end
       end
 
-      specify do
-        expect(CountriesIndex::Country.root.compose(country_with_cities)).to eq('id' => 1, 'cities' => [
-          {'id' => 1, 'name' => 'City1'}, {'id' => 2, 'name' => 'City2'}
-        ])
+      context 'parental and children fields with ignore_blank flag' do
```
rabotyaga, line 481: Is it a parent/child or nested field? The first one reminds me of https://www.elastic.co/guide/en/elasticsearch/reference/6.8/parent-join.html

spec/chewy/fields/base_spec.rb
Vitalina-Vakulchyk: I created a special section for ignore_blank option. Thank you

spec/chewy/fields/base_spec.rb
Vitalina-Vakulchyk: Blank. Thank you

spec/chewy/fields/base_spec.rb
Vitalina-Vakulchyk: I think, `nested field`.

Vitalina-Vakulchyk:
@rabotyaga @barthez 
Thank you for the comments! Please take a look again.

.rubocop_todo.yml
```diff
@@ -42,7 +41,19 @@ Lint/MissingSuper:
     - 'lib/chewy/type/adapter/object.rb'
     - 'lib/chewy/type/adapter/orm.rb'
 
-# Offense count: 36
+# Offense count: 3
+Lint/SymbolConversion:
```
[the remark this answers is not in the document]
Vitalina-Vakulchyk: @AlexVPopov I removed all the new offences. And I added `Ruby 2.7` as target version to `rubocop.yml` to avoid Rubocop CI error https://github.com/toptal/chewy/runs/2235043027?check_suite_focus=true . 

Then I asked in #chewy chat, is it okay. So, if adding target version is okay, PR will be ready for your second check :)

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ### Changes
 
+  * [#778](https://github.com/toptal/chewy/pull/778): Add `ignore_blank` option to `field` method ([@Vitalina-Vakulchyk][]):
```
rabotyaga, line 9:
```suggestion
  * [#778](https://github.com/toptal/chewy/pull/778): Add `ignore_blank` option to the `field` method ([@Vitalina-Vakulchyk][]):
```

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
...
+    * `true` by default for `geo_point` type
```
rabotyaga, line 10:
```suggestion
    * `true` by default for the `geo_point` type
```

README.md
```diff
@@ -660,6 +662,25 @@ By default, when you perform import Chewy checks whether an index exists and cre
 You can turn off this feature to decrease Elasticsearch hits count.
 To do so you need to set `skip_index_creation_on_import` parameter to `false` in your `config/chewy.yml`
 
+### Skip record fields during import
+
+You can use `ignore_blank: true` to skip fields that return `true` for `.blank?` method:
```
rabotyaga, line 667:
```suggestion
You can use `ignore_blank: true` to skip fields that return `true` for the `.blank?` method:
```

spec/chewy/fields/base_spec.rb
```diff
@@ -400,6 +400,313 @@
...
+        stub_model(:city)
+        stub_model(:country)
+
+        City.belongs_to :country
+        Location.belongs_to :city
+        City.has_one :location, -> { order :id }
+        Country.has_many :cities, -> { order :id }
+      end
+
+      context 'text fields type with and without ignore_blank option' do
```
rabotyaga, line 415:
```suggestion
      context 'text fields with and without ignore_blank option' do
```

spec/chewy/fields/base_spec.rb
```diff
@@ -400,6 +400,313 @@
...
+
+        specify do
+          expect(CountriesIndex::Country.root.compose(country_with_cities)).to eq(
+            'id' => 1, 'cities' => [
+              {'id' => 1}
+            ]
+          )
+        end
+      end
+
```
rabotyaga, line 480: Does this context add some cases, not handled by the previous one? I couldn't find any.
```suggestion
```

spec/chewy/fields/base_spec.rb
```diff
@@ -400,6 +400,313 @@
...
+          let(:country_with_cities) { Country.create!(id: 1) }
+
+          specify do
+            expect(CountriesIndex::Country.root.compose(country_with_cities))
+              .to eq('id' => 1, 'cities' => [])
+          end
+        end
+
+        context 'without ignore_blank: true flag' do
```
rabotyaga, line 540:
```suggestion
        context 'without ignore_blank option' do
```

spec/chewy/fields/base_spec.rb
```diff
@@ -400,6 +400,313 @@
...
+
+          specify do
+            expect(CountriesIndex::Country.root.compose(country_with_cities))
+              .to eq('id' => 1, 'cities' => [])
+          end
+        end
+      end
+
+      # rubocop:disable Lint/SymbolConversion
+      context 'pass ignore_blank with string key' do
```
rabotyaga, line 565: Why should we test string keys?

spec/chewy/fields/base_spec.rb
```diff
@@ -400,6 +400,313 @@
...
+            ).to eq(
+              'id' => 1, 'cities' => [
+                {'id' => 1, 'name' => 'City1'},
+                {'id' => 2, 'name' => 'City2'}
+              ]
+            )
+          end
+        end
+
+        context 'without ignore_blank: true flag' do
```
rabotyaga, line 638:
```suggestion
        context 'without ignore_blank option' do
```

rabotyaga (approved):

spec/chewy/fields/base_spec.rb
Vitalina-Vakulchyk: Just because firstly I added to #compose the check of 
`@options[:type].to_sym == :geo_point to allow users to pass types as string.` according to the next comment https://github.com/toptal/chewy/pull/778#discussion_r603470106

But after I removed this check since https://github.com/toptal/chewy/pull/778#discussion_r604161147

Do you think I can remove this specs now as unnecessary? @rabotyaga

rabotyaga: Yeah, I think it's redundant.

AlexVPopov (approved):

Vitalina-Vakulchyk:
Ruby 2.5 support was dropped in this PR according to discussion in team Slack chat https://toptal-core.slack.com/archives/C6C64B7H8/p1617365634490100

barthez (approved):

PR #787, opened 2021-04-15
[Fix #469] Print log progress on import rake task

### LINKED ISSUE #469: Allow printing of import progress during `rake chewy:reset`

Vitalina-Vakulchyk:
Allow printing of import progress during `rake chewy:reset` and `rake chewy:parallel:reset`.

* [x] Updated tests.


lib/chewy/type/adapter/object.rb
```diff
@@ -193,6 +193,16 @@ def load(ids, **options)
...
         end
 
+        def import_count(*args)
+          collection = if args.first.empty? && @target.respond_to?(import_all_method)
+            @target.send(import_all_method)
+          else
+            args.flatten(1).compact
+          end
+
+          collection.count
```
barthez, line 203: Won't it be a "bit" inefficient for huge data sources?

lib/chewy/type/import.rb
```diff
@@ -150,8 +151,10 @@ def empty_objects_or_scope?(objects_or_scope)
 
         def import_linear(objects, routine)
           ActiveSupport::Notifications.instrument 'import_objects.chewy', type: self do |payload|
+            progressbar = ProgressBar.create total: adapter.import_count(objects) if routine.options[:progressbar]
             adapter.import(*objects, routine.options) do |action_objects|
               routine.process(**action_objects)
+              progressbar.progress += action_objects.values.flatten.length if routine.options[:progressbar]
```
barthez, line 157: maybe? `flatten` is bad, maybe `flatten(1)` instead?
```suggestion
              progressbar.progress += action_objects.map { _, v| v.size }.sum if routine.options[:progressbar]
```

If you chose to apply it, do it in other places too.

Vitalina-Vakulchyk: Thank you! I applyed this change. 

Is this approach better because it is more explicit? @barthez

lib/chewy/type/adapter/object.rb
dalthon: Actually is hard to tell the advantages/disadvantages, because for huge data sources it could be particularly more interesting having a progress bar instead of just waiting without no feedback.

But yes, counting could take some time... But at least this is only calculated if someone asks for a progress bar.

WDYT? Keep as this or do you have other suggestion?

lib/chewy/type/import.rb
```diff
@@ -150,8 +151,10 @@ def empty_objects_or_scope?(objects_or_scope)
 
         def import_linear(objects, routine)
           ActiveSupport::Notifications.instrument 'import_objects.chewy', type: self do |payload|
+            progressbar = ProgressBar.create total: adapter.import_count(objects) if routine.options[:progressbar]
```
dalthon, line 154: It would be nice to add a test case that mocks this progressbar object and make sure it has its progress incremented many times as expected

lib/chewy/rake_helper.rb
```diff
@@ -303,7 +303,7 @@ def human_duration(seconds)
 
       def reset_one(index, output, parallel: false)
         output.puts "Resetting #{index}"
-        index.reset!((Time.now.to_f * 1000).round, parallel: parallel)
+        index.reset!((Time.now.to_f * 1000).round, parallel: parallel, progressbar: ENV['PROGRESSBAR'])
```
dalthon, line 306: It is nice to check content of progressbar env, because if someone pass `PROGRESSBAR=false`, here we would interpret this value `"false"` as true because it is just a string.

Also I think that `PROGRESS` might be a better shorter namefor this variable, but I have no stronger opinions about it, keep yours it if you prefer it.

dalthon (commented):
LGTM, just asked for a few tests, if it doesn't bother you much

lib/chewy/type/adapter/object.rb
Vitalina-Vakulchyk: Maybe, somebody will feel that calculations take too much time and then will turn off the progressbar.

Or does it makes sence to count the seconds of  `collection.count` executing? And we can stop the process via set `progressbar.title` to `Too many elements to output progressbar` in case of too long duration (like 2 seconds)?
@barthez @dalthon

Vitalina-Vakulchyk: @barthez 
Should I change the approuch and just output the count of imported records to avoid `collection.count`? Or any other way?

@dalthon  
And I can try to use `Mutex` to synchronize the count of records to output during parallel import.

dalthon: I am totally ok with any of those approaches:
* Using the progress bar not as default so coping with the possibility of this performance cost happens with some kind of consent
* Changing this to a simple count of elements imported, of course handling the concurrency edge cases
* Having progress bar or count depending on env parameters, having simple count as default

Since @barthez doesn't seem to much fan of this introduced extra query for count, I guess the most promising of those 3 points are the 2nd and 3rd.

 Between those 2 last, I prefer the 2nd due to its simplicity and it would not add any new dependency to the project

Vitalina-Vakulchyk: I will try to use thread to increment `progressbar.total` eventually during parallel import executing.

Vitalina-Vakulchyk:
@barthez Bartek, please take a look again on the PR.

Gemfile
```diff
@@ -12,6 +12,8 @@ gem 'kaminari-core', require: false
 gem 'parallel', require: false
 gem 'ruby-progressbar', require: false
 
+gem 'pry'
```
dalthon, line 15: Need to remove it

gemfiles/rails.6.0.activerecord.gemfile
```diff
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+gem 'pry'
```
dalthon, line 3: Remove this as well

spec/support/active_record.rb
```diff
@@ -1,6 +1,6 @@
 require 'database_cleaner'
 
-ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: 'file::memory:?cache=shared', pool: 10)
+ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: 'file::memory:?cache=shared', pool: 1)
```
dalthon, line 3: It me be interesting to explain here that test suite doesn't run properly with more than one database connection, otherwise it does not find tables

dalthon (approved):
LGTM!
Just added a few minor comments

barthez (approved):

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@
         * in rake tasks output (e.g. `Imported CitiesIndex::City in 1s, stats: index 3` -> `Imported CitiesIndex in 1s, stats: index 3`)
     * Use index name instead of type name in loader additional scope
         * e.g. `CitiesIndex.filter(...).load(city: {scope: City.where(...)})` -> `CitiesIndex.filter(...).load(cities: {scope: City.where(...)})`
+  * [#469](https://github.com/toptal/chewy/issues/469): Add ability to output progressbar with `ENV['PROGRESS']` during `reset` rake tasks ([@Vitalina-Vakulchyk][]):
+    * for `rake chewy:reset` and `rake chewy:parallel:reset`
+    * progressbar is hidden by default, set `ENV['PROGRESS']` to `true` to display it
```
rabotyaga, line 38: This should have been in the `unreleased` section, between lines 7 and 9. Or probably better in the new features sections (5-7).

rabotyaga (commented):
I think merge was a little bit hasty.
Mentioning the feature in the README would be nice.
Why did we switch import parallelization from a process-based to a thread-based one?

Vitalina-Vakulchyk:
@rabotyaga Because `Variables are protected from change` for processes in `Parallel` gem:
https://github.com/grosser/parallel

Vitalina-Vakulchyk:
@rabotyaga Thank you
I will create a PR to update `CHANGELOG` and `Readme` files.

PR #789, opened 2021-04-21
fix a typo in readme

kirillshamardin:
Fixed a typo in readme for chewy index filename user_index.rb which gives an error. The right filename is users_index.rb


rabotyaga (approved):

rabotyaga:
Thanks @kirillshamardin ! 🙇

PR #792, opened 2021-05-03
Prepare 7.0.1 release with skipped ES cluster version memoization for search requests

rabotyaga:
The “total hits” counter is an integer for ES versions < 7 and an object (hash) for the versions starting from 7.0.0. Elasticsearch added a special option, `rest_total_hits_as_int`, to ease the upgrade, that could be appended to any request and results in the old “total hits” format. Unfortunately, this option is not recognized by ES versions prior to 7.0.0, which means that we have to check the version to decide if we need this option.

Normally Chewy does memoization of the current ES version, but this might be inappropriate for the upgrade, as the version changes live.

Release the v7.0.1 Chewy version to allow dynamically detect the current Elasticsearch version.
More details in the Migration Guide.

-----------------

* (n/a) Commit message starts with `[Fix #issue-number]` (if the related issue exists).
* (n/a) Added tests.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -1,0 +1,94 @@
+name: CI
+
+on: [push]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [2.6, 2.7]
+        gemfile: [rails.5.2.activerecord, rails.6.0.activerecord, rails.6.1.activerecord]
+    name: ${{ matrix.ruby }}-${{ matrix.gemfile }}
+
+    env:
+      BUNDLE_GEMFILE: gemfiles/${{ matrix.gemfile }}.gemfile
+
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


dalthon:
@rabotyaga, don't you think that it would be nice to keep this memoization by default and only disable it with some configuration?
The issue you described seem to be a very rare case that in order to be handled puts an extra burden in all requests

rabotyaga:
> @rabotyaga, don't you think that it would be nice to keep this memoization by default and only disable it with some configuration?
> The issue you described seem to be a very rare case that in order to be handled puts an extra burden in all requests

@dalthon You're absolutely right. And this configuration is release `7.0.1`. Please take a look at https://github.com/toptal/chewy/pull/792/files#diff-134d08723ee8d2712d2f3250f8936bda26c620b022ec5d7003b3c06b114938c0R20.

dalthon (approved):

adamors (approved):

rabotyaga:
We don't need to actually merge this, the release is published from the tag: https://github.com/toptal/chewy/releases/tag/v7.0.1

PR #798, opened 2021-05-19
[Fix #795] Change implementation of Order param

### LINKED ISSUE #795: cannot have more than 1 script sorting
In elasticsearch, you can have as many script sorting as you want.
For example, there is a test index with two keyword fields "region" and "country", to sort by region and country alphabetically, by using script sorting you can do 
```
{
  "sort": [
    {
      "_script": {
        "type": "string",
        "script": {
          "inline": "doc['region'].value"
        },
        "order": "desc"
      }
    },
    {
      "_script": {
        "type": "string",
        "script": {
          "inline": "doc['country'].value"
        },
        "order": "desc"
      }
    }
  ]
}
```
(In real world, you can simply do `{ "sort": ["region", "country"] }`,  this example is just to explain the issue)
However, when I use the order method of chewy I can't add more than 1 script sorting
```
TestsIndex.order([
    {
      "_script": {
        "type": "string",
        "script": {
          "inline": "doc['region'].value"
        },
        "order": "desc"
      }
    },
    {
      "_script": {
        "type": "string",
        "script": {
          "inline": "doc['country'].value"
        },
        "order": "desc"
      }
    }
  ])
```
it returns
```
<TestsIndex::Query {:index=>["tests"], :type=>["user"], :body=>{:sort=>[{"_script"=>{:type=>"string", :script=>{:inline=>"doc['country'].value"}, :order=>"desc"}}]}}>
```
The region script sorting part is gone.

--------

## Expected behavior

Expect query to be like
```
<TestsIndex::Query {:index=>["tests"], :type=>["user"], :body=>{:sort=>[{"_script"=>{:type=>"string", :script=>{:inline=>"doc['region'].value"}, :order=>"desc"}}, {"_script"=>{:type=>"string", :script=>{:inline=>"doc['country'].value"}, :order=>"desc"}}]}}>
```
## Actual behavior
Actual query is 
```
<TestsIndex::Query {:index=>["tests"], :type=>["user"], :body=>{:sort=>[{"_script"=>{:type=>"string", :script=>{:inline=>"doc['country'].value"}, :order=>"desc"}}]}}>
```
## Steps to reproduce the problem

Define index
```
class TestsIndex < Chewy::Index
  define_type User do
    field :region, type: :keyword
    field :country, type: :keyword
  end
end
```
Then in console do
```
TestsIndex.order([
    {
      "_script": {
        "type": "string",
        "script": {
          "inline": "doc['region'].value"
        },
        "order": "desc"
      }
    },
    {
      "_script": {
        "type": "string",
        "script": {
          "inline": "doc['country'].value"
        },
        "order": "desc"
      }
    }
  ])
```

## Version Information
Chewy 5.0.0 and 6.0.0 and 7.0.0
Elasticsearch 6.4
ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-darwin19]
Rails 5.1.7

--------
By looking into the source code, I found that the value of Chewy::Search::Parameters::Order is a hash and the issue is because the key of script sorting must always be "_script".
```
=> #<Chewy::Search::Parameters:0x00007fa76fb3c950
 @storages=
  {:order=>#<Chewy::Search::Parameters::Order:0x00007fa76fb3c6a8 @value={"_script"=>{:type=>"string", :script=>{:inline=>"doc['country'].value"}, :order=>"desc"}}>,
   :types=>#<Chewy::Search::Parameters::Types:0x00007fa76fb4f438 @value=[]>,
   :none=>#<Chewy::Search::Parameters::None:0x00007fa76fb4ea60 @value=false>,
   :filter=>#<Chewy::Search::Parameters::Filter:0x00007fa76fb4ea38 @value=#<Chewy::Search::Parameters::QueryStorage::Bool:0x00007fa76fb4e7e0 @minimum_should_match=nil, @must=[], @must_not=[], @should=[]>>,
   :query=>#<Chewy::Search::Parameters::Query:0x00007fa76fb4e3d0 @value=#<Chewy::Search::Parameters::QueryStorage::Bool:0x00007fa76fb4e1c8 @minimum_should_match=nil, @must=[], @must_not=[], @should=[]>>}>
```

jiajiawang:
Use Array to store the value of Chewy::Search::Parameters::Order.
To allow multiple sorting options that may have the same key name.
For example script based sorting whose key will always be `_script`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@
 
 ### Changes
 
+  * [#795](https://github.com/toptal/chewy/issues/795): **(Breaking)** Change the Chewy::Search::Parameters::Order implementation to use Array ([@jiajiawang][]):
+    * To allow multiple sorting options that may have the same key name. For example script based sorting whose key will always be `_script`.
+    * Behaviour change of chained `order` calls.
+      * e.g. `.order(_script: {a: 1}).order(_script: {b: 2})`
+        * Before `{:sort=>[{"_script"=>{:b=>2}}]}`
+        * After  `{:sort=>[{"_script"=>{:a=>1}},{"_script"=>{:b=>2}}]}`
+
   * [#654](https://github.com/toptal/chewy/issues/654): Add helpers and matchers for testing ([@Vitalina-Vakulchyk][]):
     * `mock_elasticsearch_response` helpers both Rspec and Minitest - to mock elasticsearch response
     * `mock_elasticsearch_response_sources` helpers both Rspec and Minitest - to mock elasticsearch response sources
@@ -13,9 +20,6 @@
     * `build_query` matcher for Rspec - to compare request and expected query (returns `true`/`false`)
 
 ### Bugs Fixed
-
-  * [#795](https://github.com/toptal/chewy/issues/795): Change the Chewy::Search::Parameters::Order implementation to use Array ([@jiajiawang][]):
-    * To allow multiple sorting options that may have the same key name. For example script based sorting whose key will always be `_script`.
 
 ## 7.2.1 (2021-05-11)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 
 ### Bugs Fixed
 
+  * [#795](https://github.com/toptal/chewy/issues/795): Change the Chewy::Search::Parameters::Order implementation to use Array ([@jiajiawang][]):
+    * To allow multiple sorting options that may have the same key name. For example script based sorting whose key will always be `_script`.
+
 ## 7.2.1 (2021-05-11)
 
 ### New Features
@@ -637,6 +640,7 @@
 [@inbeom]: https://github.com/inbeom
 [@jesjos]: https://github.com/jesjos
 [@JF-Lalonde]: https://github.com/JF-Lalonde
+[@jiajiawang]: https://github.com/jiajiawang
 [@jimmybaker]: https://github.com/jimmybaker
 [@jirikolarik]: https://github.com/jirikolarik
 [@jirutka]: https://github.com/jirutka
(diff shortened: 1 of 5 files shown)


rabotyaga:
Hey @jiajiawang 👋 !
This is really nice!
There is one thing, that bothers me. Actually, this is a breaking change. Someone might use chaining of request methods to build complex queries and they might chain `order` method too. After this change chained queries like `SomeIndex.order('name').order('name'=>{'order' => 'desc'})` will start work differently.
Also, there is no much sense in having different sort options for the same field (`{:sort=>["name", {"name"=>{"order"=>"desc"}}]}`), except the special `_script` field.

What if we keep the previous behavior for all keys, except the `_script` one?

jiajiawang:
Hi @rabotyaga 👋 !
Thanks for the quick response.
You right. It is a breaking change.
Though, `_script` isn't the only problem we have. Similarly there is also `_geo_distance`.
Even for sorting on one certain field multiple times, there are also legitimate use cases.

For example, a numeric field `price` whose value is `[10, 20, 30, ...]` (price for different SKUs).
We may want to do `{sort: [{price: { mode: 'min' }}, {price: {mode: 'avg'}}]}` - sort products by minium SKU price, then by average SKU price if minimum is the same.

Most importantly I feel we should accept the sort body as long as it can be accepted by ES even though they sometime may don't make sense.

rabotyaga:
Valid points.
Let's then mark this change as breaking in the CHANGELOG (https://github.com/toptal/chewy/blob/master/CONTRIBUTING.md#changelog-entry-format) and somehow emphasize, that chained `order` calls with the same keys now work differently.

jiajiawang:
CHNAGELOG updated.

rabotyaga:
@jiajiawang Thank you very much!

PR #807, opened 2021-07-19
Use thread-local variables instead of fiber-local variables to store global configuration.

bolshakov:
Use thread-local variables instead of fiber-local variables to store global configuration.

1. Currently, we’re using fiber-local variables. This is indistinguishable from the pre-ruby-1.9 thread-local variables, unless you use fibers and use chewy in those fibers. But if you do - all fiber-local variables are extinguished inside any fiber, explicitly created by `Fiber.new`.
2. If we switch to thread-local variables - i.e. `thread_variable_get/thread_variable_set` - we’re fixing the issue with variables presence inside explicitly created fibers, but introducing a potential issue of having a mess between fibers of one thread, if one decides to use different clients or different strategies in different fibers.
3. If we continue to use fiber-local variables and add a patch on the dry-effects side this fixes the [issue with the dry-effects](https://github.com/dry-rb/dry-effects/issues/82), but a more general case is left unhandled, and, what sounds particularly bad to me, we’re introducing coupling between 2 completely unrelated gems.

As we discussed with @kml and @rabotyaga, we're changing fiber-local to thread-local for now because it's unlikely it may affect anyone using chewy in any negative way.  If this rare use case bubbles up later - we can address it by introducing a configuration option. 

-----------------

* [x] Added tests.

### DIFF
--- a/lib/chewy.rb
+++ b/lib/chewy.rb
@@ -65,20 +65,12 @@ module Chewy
     # A thread-local variables accessor
     # @return [Hash]
     def current
-      initialize_current_storage
-
-      current
-    end
-
-    # @api private
-    def initialize_current_storage
-      Thread.current.thread_variable_set(:chewy, {})
-
-      define_singleton_method(:current) do
-        Thread.current.thread_variable_get(:chewy)
+      unless Thread.current.thread_variable?(:chewy)
+        Thread.current.thread_variable_set(:chewy, {})
       end
+
+      Thread.current.thread_variable_get(:chewy)
     end
-    private :initialize_current_storage
 
     # Derives an index for the passed string identifier if possible.
     #
--- a/lib/chewy.rb
+++ b/lib/chewy.rb
@@ -65,8 +65,20 @@ module Chewy
     # A thread-local variables accessor
     # @return [Hash]
     def current
-      Thread.current[:chewy] ||= {}
+      initialize_current_storage
+
+      current
     end
+
+    # @api private
+    def initialize_current_storage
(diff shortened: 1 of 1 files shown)


lib/chewy.rb
```diff
@@ -65,8 +65,20 @@ class << self
...
+      initialize_current_storage
+
+      current
+    end
+
+    # @api private
+    def initialize_current_storage
+      Thread.current.thread_variable_set(:chewy, {})
+
+      define_singleton_method(:current) do
```
kml, line 77: I think it would be less complex with `def current` doing it all without  dynamic overriding method after first use.
https://www.rubydoc.info/stdlib/core/Thread#thread_variable%3F-instance_method to check if variable is set and set default if it's not would be fine.

bolshakov: C'mon, it's just a few lines of code and a pretty common trick. Are there any drawbacks of such an approach?

kml: I think there is even other issue - in case of two threads first one will call `thread_variable_set`,
but all other threads will already use overriden method that only `_get`.
This way in multi-threaded environment access to `Chewy.current` from other thread than first one would end up with:
```
irb(main):002:0> Thread.current.thread_variable_get(:chewy)
=> nil
```

(as Thread.current is relative to the caller; it would be issue-less only in case of Thread.main)

bolshakov: Good point @kml

kml (approved):
LGTM

rabotyaga (approved):

PR #827, opened 2021-12-10
LazySidekiq strategy

sl4vr:
This PR implements proof of concept for new Chewy strategy. The strategy allows to avoid synchronous evaluation of index queries. Instead all the callbacks will be saved in class attribute and model type and id will be stashed into strategy instance. When strategy block ends it will enqueue `IndicesUpdateWorker` Sidekiq job with serialized stash as argument. The job will load records and run all the callbacks within `sidekiq` strategy block, which in turn will enqueue Sidekiq jobs updating indices.
Since we cannot schedule `IndicesUpdateWorker` job for destroyed records, instead all the callbacks will be run synchronously and resulted index types and ids will be stashed into strategy instance and run on the `lazy_sidekiq` strategy block end like if they would be run on the end of `sidekiq` strategy block.

Some concerns to consider:
* One cannot rely on state within `update_index` blocks because record will be reinstantiated on the time when callbacks will be run. We can however serialize `changes` and `previous_changes` to Sidekiq jobs, or even specify what to serialize.

-----------------

* [x] Added tests.


lib/chewy/strategy/lazy_sidekiq.rb
```diff
@@ -0,0 +1,54 @@
...
+
+      def leave
+        @stash.each do |model_name, id|
+          ::Sidekiq::Client.push(
+            'queue' => sidekiq_queue,
+            'class' => Chewy::Strategy::LazySidekiq::LazyWorker,
+            'args'  => [model_name, id]
+          )
+        end
+      end
```
mrzasa, line 47: What happens if we update multiple objects? Do we run callback for them in separate jobs? If it's the case let's consider the situation when the collection is updated when callback are run. We may end up with some of the objects indexed with the intended state (with the change that scheduled the indexing) and some of them - with the new state. If we index with every update it's not a problem, but if we index only some updates, we'll end up with inconsistent state in  elastic.

Example:
```
Chewy.strategy(:lazy_sidekiq) do
  User.find(1).update!(score: 100, name: 'John')
  User.find(2).update!(score: 200, name: 'Luke')  
  # We're scheduling lazy sidekiq jobs here
end

# When the jobs are running, we're updating  records without sidekiq
 User.find(1).update!(score: 500)  
 User.find(2).update!(score: 400)  
 User.find(3).update!(score: 200)  
```
It may happen that we index
```
{`name`: `John`, 'score': 100}
{`name`: `Luke`, 'score': 400}
```
I'm not sure if we should care about this scenario, but I wanted to point think it's good to at  least think about it before merging.

sl4vr: @mrzasa Currently in this strategy we schedule different jobs for every updated record within strategy block. And then when the job is performed it schedules a job for every callback updating index.

I think I got your concern. We could schedule only one job in the end of the strategy block with an array of type/id tuples of records that should be updated, .e.g. `[['User', 1], ['User', 2]]`. Then when the job run we can run all the chewy callbacks on records we have and schedule a new job for every index, just like we do in `Sidekiq` strategy. Would it solve that possible issue?

mrzasa: Yes, we're thinking about the same issue and yes- I think that scheduling only one job should help.

spec/chewy/strategy/lazy_sidekiq_spec.rb
```diff
@@ -0,0 +1,192 @@
...
+  describe Chewy::Strategy::LazySidekiq do
+    around do |example|
+      sidekiq_settings = Chewy.settings[:sidekiq]
+      Chewy.settings[:sidekiq] = {queue: 'low'}
+      Chewy.strategy(:lazy_sidekiq) { example.run }
+      Chewy.settings[:sidekiq] = sidekiq_settings
+    end
+    before { ::Sidekiq::Worker.clear_all }
+
+    context 'tests from sidekiq_spec + some more' do
```
sl4vr, line 15: Specs will be rewritten and probably expanded if the rest is ok

lib/chewy/index/observe/active_record_methods.rb
```diff
@@ -0,0 +1,87 @@
...
+              after_destroy :run_chewy_callbacks
+            end
+          end
+
+          def update_index(type_name, *args, &block)
+            callback_options = Observe.extract_callback_options!(args)
+            update_proc = Observe.update_proc(type_name, *args, &block)
+            callback = Chewy::Index::Observe::Callback.new(update_proc, callback_options)
+
+            initialize_chewy_callbacks if chewy_callbacks.empty?
```
sl4vr, line 79: I would love to define AR callbacks in `included` block however when `ActiveRecord::Base` includes `ActiveRecordMethods` Chewy config is even not initialized yet. And even if we wait for Chewy config to initialize - `Chewy.use_after_commit_callbacks` will be always true because it can be overwritten in initializers, which are being executed later on.
So here we hack and define AR callbacks just when first chewy indexing callback is deined.

lib/chewy/strategy/lazy_sidekiq.rb
sl4vr: @mrzasa should be fixed now

rabotyaga (approved):

spec/chewy/strategy/lazy_sidekiq_spec.rb
```diff
@@ -0,0 +1,230 @@
...
+
+        context 'city updates' do
+          let!(:country1) { Country.create!(id: 1) }
+          let!(:country2) { Country.create!(id: 2) }
+          let!(:city) { City.create!(id: 1, country: country1) }
+
+          it 'does not update index of removed entity because model state on the moment of save cannot be fetched' do
+            expect { city.update!(country: nil) }.not_to update_index('countries', strategy: :lazy_sidekiq)
+          end
+          it 'does not update index of removed entity because model state on the moment of save cannot be fetched' do
```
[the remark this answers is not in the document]
sl4vr: It does match, but probably the description is just confusing?
We had `city` with `country1` and we update it with `country2`, it will reindex `country2` only.
Similar test from any other strategy that keeps the state would look like:
```ruby
expect { city.update!(country: country2) }.to update_index('countries', strategy: :lazy_sidekiq).and_reindex(country1, country2).only
```

spec/chewy/strategy/lazy_sidekiq_spec.rb
```diff
@@ -0,0 +1,230 @@
...
+            attr_accessor :state_dependent_update_condition
+          end
+        end
+
+        context 'city updates' do
+          let!(:country1) { Country.create!(id: 1) }
+          let!(:country2) { Country.create!(id: 2) }
+          let!(:city) { City.create!(id: 1, country: country1) }
+
+          it 'does not update index of removed entity because model state on the moment of save cannot be fetched' do
```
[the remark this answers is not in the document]
rabotyaga: Yeah, that's what I was referring to as `model id`. I thought the test name echoes `state_dependent_update_condition` somehow.

spec/chewy/strategy/lazy_sidekiq_spec.rb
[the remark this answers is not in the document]
rabotyaga: I see, thanks. Maybe confusing a bit. I thought test name echoes `state_dependent_update_condition` somehow.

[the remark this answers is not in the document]
sl4vr: Yeah, the only test that uses `state_dependent_update_condition` is `does not update index because state of attribute cannot be fetched` and it demonstrates that if condition is state dependent index won't be updated.
I'll try to come up with more descriptive context and example names.

mrzasa (approved):

lib/chewy/index/observe/callback.rb
```diff
@@ -0,0 +1,34 @@
...
+      class Callback
+        def initialize(executable, filters = {})
+          @executable = executable
+          @if_filter = filters[:if]
+          @unless_filter = filters[:unless]
+        end
+
+        def call(context)
+          return if !@if_filter.nil? && !eval_filter(@if_filter, context)
+          return if !@unless_filter.nil? && eval_filter(@unless_filter, context)
```
riffraff, line 13: how about adding 
```
when nil
  true
```
in `eval_filter`? 
Then this could become
```suggestion
          return unless eval_filter(@if_filter, context)
          return if eval_filter(@unless_filter, context)
```
whih is nicely symmetrical :D

riffraff: argh, you're right of course

lib/chewy/strategy/lazy_sidekiq.rb
```diff
@@ -0,0 +1,58 @@
...
+    #     Post.all.map(&:save) # And here
+    #     # It schedules import of all the changed users and posts right here
+    #   end
+    #
+    class LazySidekiq < Sidekiq
+      class IndicesUpdateWorker
+        include ::Sidekiq::Worker
+
+        def perform(models)
+          Chewy.strategy(:sidekiq) do
```
barthez, line 18: What it result in one more job scheduled? Wouldn't it be better to use `urgent` strategy here?

sl4vr: @barthez, good point. Initially I made it intuitively as an offspring of `sidekiq` strategy. Actually I don't have a strong opinion on which strategy to use in `LazySidekiq::IndicesUpdateWorker`, the only thing that comes up to my mind is that `sidekiq` can be more performant due to multiple threads.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### New Features
 
+ * [#827](https://github.com/toptal/chewy/pull/827): Add `:lazy_sidekiq` strategy, that defers not only importing but also `update_index` callback evaluation for created and update objects. ([@sl4vr][])
```
rabotyaga, line 7:
```suggestion
 * [#827](https://github.com/toptal/chewy/pull/827): Add `:lazy_sidekiq` strategy, that defers not only importing but also `update_index` callback evaluation for created and updated objects. ([@sl4vr][])
```

rabotyaga (approved):

mrzasa (approved):

PR #839, opened 2022-03-09
Fix Chewy::Search::Request#only and Chewy::Search::Request#except docs example

suke:


rabotyaga:
Hey @suke 👋 !
Thank you very much!

PR #848, opened 2022-06-09
Fix incorrect pagination in journaling

konalegi:
This PR fixes two bugs: 
* Removes the retries principle, just because `Chewy::Stash::Journal.entries` methods as any es query have a limit of default 10 and in case you have more than 100 they are going to be ignored. Instead, pagination is based on the `total_count` will scan through all entires in the journal.
* Another bug is a result of the assumption that entries in the query are sorted by `created_at`, but in reality, it's not.
As an example imagine the following entries in the journal 

```
1. Entry (id: 1, created_at: '12:01:00')
2. Entry (id: 3, created_at: '12:04:00')
3. Entry (id: 2, created_at: '12:03:00')
4. Entry (id: 4, created_at: '12:05:00')
```

and the current algorithm step by step 
1. fetch 2 items older than `12:00:00`, as a result, entries with ids  1, 3 will be fetched ([src](https://github.com/toptal/chewy/blob/master/lib/chewy/journal.rb#L26))
2. calculate the latest in the batch ([src](https://github.com/toptal/chewy/blob/master/lib/chewy/journal.rb#L34)), latest created_at will be `12:04:00`
3. fetch 2 items older than`12:04:00`, as a result, we are going have entry with `id: 4` only 
4. Entry with 2 will be lost 

This bug fix is simple, fetch all entries ordered by `created_at`

-----------------

* [x] Added tests.

### DIFF
--- a/lib/chewy/journal.rb
+++ b/lib/chewy/journal.rb
@@ -16,14 +16,17 @@ module Chewy
     # specified indexes.
     #
     # @param since_time [Time, DateTime] timestamp from which changes will be applied
-    # @param retries [Integer] maximum number of attempts to make journal empty, 10 by default
+    # @param fetch_limit [Int] amount of entries to be fetched on each cycle
     # @return [Integer] the amount of journal entries found
-    def apply(since_time, retries: 10, **import_options)
+    def apply(since_time, fetch_limit: 10, **import_options)
       stage = 1
       since_time -= 1
       count = 0
-      while stage <= retries
-        entries = Chewy::Stash::Journal.entries(since_time, only: @only).to_a.presence or break
+
+      total_count = entries(since_time, fetch_limit).total_count
+
+      while count < total_count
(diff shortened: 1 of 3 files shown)


barthez (approved):

lib/chewy/journal.rb
```diff
@@ -16,14 +16,16 @@ def initialize(*only)
     # specified indexes.
     #
     # @param since_time [Time, DateTime] timestamp from which changes will be applied
-    # @param retries [Integer] maximum number of attempts to make journal empty, 10 by default
     # @return [Integer] the amount of journal entries found
-    def apply(since_time, retries: 10, **import_options)
+    def apply(since_time, fetch_limit: 10, **import_options)
```
mrzasa, line 20: I'd document the `fetch_limit` param in the method comment.

mrzasa (approved):

PR #857, opened 2022-09-19
Allow asynchronously cleanup journal

konalegi:
For the cases when we need to remove a large journal we could benefit from using the async execution provided by [delete-by-query](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html#docs-delete-by-query-task-api) methods

Usage example 
```
rake chewy:journal:clean -- --wait_for_completion=false --requests_per_second=10 --scroll_size=5000
```

-----------------

* [x] Added tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,6 +677,8 @@ You may be wondering why do you need it? The answer is simple: not to lose the d
 
 Imagine that you reset your index in a zero-downtime manner (to separate index), and at the meantime somebody keeps updating the data frequently (to old index). So all these actions will be written to the journal index and you'll be able to apply them after index reset using the `Chewy::Journal` interface.
 
+When enabled, journal can grow to enormous size, consider setting up cron job that would clean it occasionally using [`chewy:journal:clean` rake task](#chewyjournal).
+
 ### Index manipulation
 
 ```ruby
@@ -1142,6 +1144,17 @@ This namespace contains two tasks for the journal manipulations: `chewy:journal:
 ```bash
 rake chewy:journal:apply["$(date -v-1H -u +%FT%TZ)"] # apply journaled changes for the past hour
 rake chewy:journal:apply["$(date -v-1H -u +%FT%TZ)",users] # apply journaled changes for the past hour on UsersIndex only
+```
+
+When the size of the journal becomes very large, the classical way of deletion would be obstructive and resource consuming. Fortunately, Chewy internally uses [delete-by-query](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/docs-delete-by-query.html#docs-delete-by-query-task-api) ES function which supports async execution with batching and [throttling](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html#docs-delete-by-query-throttle).
+
+The available options, which can be set by ENV variables, are listed below:
+* `WAIT_FOR_COMPLETION` - a boolean flag. It controls async execution. It waits by default. When set to `false` (`0`, `f`, `false` or `off` in any case spelling is accepted as `false`), Elasticsearch performs some preflight checks, launches the request, and returns a task reference you can use to cancel the task or get its status.
+* `REQUESTS_PER_SECOND` - float. The throttle for this request in sub-requests per second. No throttling is enforced by default.
+* `SCROLL_SIZE` - integer. The number of documents to be deleted in single sub-request. The default batch size is 1000.
+
+```bash
+rake chewy:journal:clean WAIT_FOR_COMPLETION=false REQUESTS_PER_SECOND=10 SCROLL_SIZE=5000
 ```
 
 ### RSpec integration
--- a/lib/chewy/rake_helper.rb
+++ b/lib/chewy/rake_helper.rb
@@ -19,6 +19,9 @@ module Chewy
       output.puts "  Applying journal to #{targets}, #{count} entries, stage #{payload[:stage]}"
     end
 
+    DELETE_BY_QUERY_OPTIONS = %w[WAIT_FOR_COMPLETION REQUESTS_PER_SECOND SCROLL_SIZE].freeze
+    FALSE_VALUES = %w[0 f false off].freeze
+
     class << self
       # Performs zero-downtime reindexing of all documents for the specified indexes
       #
@@ -162,7 +165,7 @@ module Chewy
 
         subscribed_task_stats(output) do
           output.puts "Applying journal entries created after #{time}"
-          count = Chewy::Journal.new(indexes_from(only: only, except: except)).apply(time)
+          count = Chewy::Journal.new(journal_indexes_from(only: only, except: except)).apply(time)
           output.puts 'No journal entries were created after the specified time' if count.zero?
         end
       end
@@ -181,12 +184,16 @@ module Chewy
       # @param except [Array<Chewy::Index, String>, Chewy::Index, String] indexes to exclude from processing
       # @param output [IO] output io for logging
       # @return [Array<Chewy::Index>] indexes that were actually updated
-      def journal_clean(time: nil, only: nil, except: nil, output: $stdout)
+      def journal_clean(time: nil, only: nil, except: nil, delete_by_query_options: {}, output: $stdout)
         subscribed_task_stats(output) do
           output.puts "Cleaning journal entries created before #{time}" if time
-          response = Chewy::Journal.new(indexes_from(only: only, except: except)).clean(time)
-          count = response['deleted'] || response['_indices']['_all']['deleted']
-          output.puts "Cleaned up #{count} journal entries"
+          response = Chewy::Journal.new(journal_indexes_from(only: only, except: except)).clean(time, delete_by_query_options: delete_by_query_options)
+          if response.key?('task')
+            output.puts "Task to cleanup the journal has been created, #{response['task']}"
+          else
+            count = response['deleted'] || response['_indices']['_all']['deleted']
+            output.puts "Cleaned up #{count} journal entries"
+          end
         end
       end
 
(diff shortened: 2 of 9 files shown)


lib/chewy/rake_helper.rb
```diff
@@ -228,6 +232,25 @@ def update_mapping(name:, output: $stdout)
...
+      # @see https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html
+      #
+      # @example
+      #   Chewy::RakeHelper.parse_delete_by_query_options(["chewy:journal:clean", "--", "--wait_for_completion=false", "--requests_per_second=10", "--scroll_size=5000"])
+      #
+      def parse_delete_by_query_options(args)
+        str_to_bool = ->(str) { {'true' => true, 'false' => false}.fetch(str) { raise('unknown value for boolean is provided') } }
+        options = {}
+        opts = OptionParser.new
+        opts.on('-w', '--wait_for_completion=ARG', String) { |val| options[:wait_for_completion] = str_to_bool.call(val) }
```
barthez, line 245: With such option, you don't need string parser.
```suggestion
        opts.on('-w', '--[no-]wait') { |val| options[:wait_for_completion] = val }
```

lib/chewy/rake_helper.rb
```diff
@@ -228,6 +232,25 @@ def update_mapping(name:, output: $stdout)
...
+        opts.on('-r', '--requests_per_second=ARG', Float) { |rps| options[:requests_per_second] = rps }
+        opts.on('-r', '--scroll_size=ARG', Integer) { |ss| options[:scroll_size] = ss }
```
barthez, line 247:
```suggestion
        opts.on('-s', '--scroll_size=ARG', Integer) { |size| options[:scroll_size] = size }
```

README.md
```diff
@@ -1144,6 +1146,17 @@ rake chewy:journal:apply["$(date -v-1H -u +%FT%TZ)"] # apply journaled changes f
 rake chewy:journal:apply["$(date -v-1H -u +%FT%TZ)",users] # apply journaled changes for the past hour on UsersIndex only
 ```
 
+For the cases when journal has grown up to enormous size, classical way of deletion is not quite possible. Fortunately chewy internally uses [delete-by-query](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/docs-delete-by-query.html#docs-delete-by-query-task-api) ES function which support async execution with batching and [throttling](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html#docs-delete-by-query-throttle).
```
AlexVPopov, line 1149:
```suggestion
When the size of the journal becomes very large, the classical way of deletion is not possible. Fortunately, Chewy internally uses [delete-by-query](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/docs-delete-by-query.html#docs-delete-by-query-task-api) ES function which supports async execution with batching and [throttling](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html#docs-delete-by-query-throttle).
```

README.md
```diff
@@ -1144,6 +1146,17 @@ rake chewy:journal:apply["$(date -v-1H -u +%FT%TZ)"] # apply journaled changes f
...
+
+Available options are listed below, they can by set by ENV variables:
```
AlexVPopov, line 1151:
```suggestion
The available options, which can be set by ENV variables, are listed below:
```

README.md
```diff
@@ -1144,6 +1146,17 @@ rake chewy:journal:apply["$(date -v-1H -u +%FT%TZ)"] # apply journaled changes f
...
+* `WAIT_FOR_COMPLETION` - boolean flag, options controls async execution, It waits by default, when set to `false`, Elasticsearch performs some preflight checks, launches the request, and returns a task you can use to cancel or get the status of the task
```
AlexVPopov, line 1152:
```suggestion
* `WAIT_FOR_COMPLETION` - a boolean flag. It controls async execution. It waits by default. When set to `false`, Elasticsearch performs some preflight checks, launches the request, and returns a task you can use to cancel the task or get its status.
```

README.md
```diff
@@ -1144,6 +1146,17 @@ rake chewy:journal:apply["$(date -v-1H -u +%FT%TZ)"] # apply journaled changes f
...
+* `REQUESTS_PER_SECOND` - float, The throttle for this request in sub-requests per second
```
AlexVPopov, line 1153:
```suggestion
* `REQUESTS_PER_SECOND` - float. The throttle for this request in sub-requests per second.
```

README.md
```diff
@@ -1144,6 +1146,17 @@ rake chewy:journal:apply["$(date -v-1H -u +%FT%TZ)"] # apply journaled changes f
...
+* `SCROLL_SIZE` - integer, Size of the scroll request that powers the operation
```
AlexVPopov, line 1154:
```suggestion
* `SCROLL_SIZE` - integer. The size of the scroll request that powers the operation.
```

AlexVPopov: I'm not quite sure what that means. :)

mrzasa: I think that we could 2-3 sentences of explaination of how the delete by query works in https://github.com/toptal/chewy/pull/857/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R1149. I'd describe what it is and how is uses the bulk API. With such an introduction it might be easier to understand all the params that we expose as ENVs. WDYT?

README.md
mrzasa: > returns a task you can use to cancel the task or get its status.

maybe

> returns a special task you can use to cancel the deletion task or get the status of the deletion task.

I find a bit confusing to have two different tasks referred in the same sentence.

mrzasa: Also, in the code we use `FALSE_VALUES = %w[0 f false off].freeze`, could we document here, how to set a `false` value?

README.md
mrzasa, line 1154: I'd also document explicitly what are the default values if we don't set the config ENV vars.

README.md
barthez: This is the same task. In the response you are getting task (reference - an id) that you can use to can use when requesting the task in Task API.

README.md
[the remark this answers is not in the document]
barthez: _Impossible is nothing_

it would make "classical way of deleting" obstructive and time consuming.

README.md
barthez: The way `delete-by-query` works in described in the link. If you don't understand the options of `delte-by-query` because you don't know how it works you reach ES documentation. I don't see much point of copying ES documentation here.

mrzasa (approved):

lib/chewy/rake_helper.rb
```diff
@@ -181,12 +184,16 @@ def journal_apply(time: nil, only: nil, except: nil, output: $stdout)
...
         subscribed_task_stats(output) do
           output.puts "Cleaning journal entries created before #{time}" if time
-          response = Chewy::Journal.new(indexes_from(only: only, except: except)).clean(time)
-          count = response['deleted'] || response['_indices']['_all']['deleted']
-          output.puts "Cleaned up #{count} journal entries"
+          response = Chewy::Journal.new(journal_indexes_from(only: only, except: except)).clean(time, delete_by_query_options: delete_by_query_options)
+          if response.key?('task')
+            output.puts "Task to cleanup the journal has been created, #{response['task']}"
+          else
+            count = response['deleted'] || response['_indices']['_all']['deleted']
```
kamtop, line 194: Here you have a variable `_indices` and above you use `indexes` as the plural version. What do you think about keeping it in one form, for consistency?

README.md
```diff
@@ -1144,6 +1146,17 @@ rake chewy:journal:apply["$(date -v-1H -u +%FT%TZ)"] # apply journaled changes f
...
+
+The available options, which can be set by ENV variables, are listed below:
+* `WAIT_FOR_COMPLETION` - a boolean flag. It controls async execution. It waits by default. When set to `false` (`0`, `f`, `false` or `off` in any case spelling is accepted as `false`), Elasticsearch performs some preflight checks, launches the request, and returns a task reference you can use to cancel the task or get its status.
+* `REQUESTS_PER_SECOND` - float. The throttle for this request in sub-requests per second. No throttling is enforced by default.
+* `SCROLL_SIZE` - integer. The number of documents to be deleted in single sub-request. The default batch size is 1000.
+
+```bash
+rake chewy:journal:clean WAIT_FOR_COMPLETION=false REQUESTS_PER_SECOND=10 SCROLL_SIZE=5000
```
kamtop, line 1157: If the `REQUESTS_PER_SECOND` is a float, maybe you should have an example with a float value not an integer.

spec/chewy/rake_helper_spec.rb
```diff
@@ -502,4 +529,45 @@
       end
     end
   end
+
+  describe '.delete_by_query_options_from_env' do
+    subject(:options) { described_class.delete_by_query_options_from_env(env) }
+    let(:env) do
+      {
+        'WAIT_FOR_COMPLETION' => 'false',
+        'REQUESTS_PER_SECOND' => '10',
```
kamtop, line 538: `REQUESTS_PER_SECOND` is defined as a float, so why do you use an integer in the example? 
Why not: `'REQUESTS_PER_SECOND' => '10.0',` ?

lib/chewy/rake_helper.rb
barthez: `response['_indices']` is the ES response, I have no control over that. `journal_indexes_from` is a method name in this helper - all other methods are using `indexes`.

kamtop: ok, so let's keep it like that

README.md
barthez: This parameter is floating point number as it is the type accepted by the API. It don't see much improvement in presenting some value that would be used very unlikely. This also shows that you can pass the value as plain integer and it will be converted to float for API call.

kamtop (approved):

PR #882, opened 2023-04-19
Fix memory leak for ruby 3.2

konalegi:
**Replace this text with a summary of the changes in your PR.
The more detailed you are, the better.**

-----------------

* [ ] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences.
* [ ] Wrote [good commit messages][1].
* [ ] Feature branch is up-to-date with `master` (if not - rebase it).
* [ ] Added tests.


lib/chewy/index/crutch.rb
```diff
@@ -12,14 +12,15 @@ class Crutches
         def initialize(index, collection)
           @index = index
           @collection = collection
+          @storage = {}
```
barthez, line 15:
```suggestion
          @crutches = {}
```

lib/chewy/index/crutch.rb
```diff
@@ -12,14 +12,15 @@ class Crutches
...
-                @#{name} ||= @index._crutches[:#{name}].call @collection
-              end
-            METHOD
+            @storage[name] = @index._crutches[:"#{name}"].call @collection
           end
         end
+
+        def fetch(name)
+          @storage[name]
+        end
```
barthez, line 23: In that case we don't break the syntax.

```suggestion
        end

        def method_missing(name, *, **)
          return self[name] if @index._crutches.key?(name)

          super
        end
        
        def respond_to_missing?(name, include_private = false)
          @index._crutches.key?(name) || super
        end

        def [](name)
          @crutches[name] ||= @index._crutches[:"#{name}"].call(@collection)
        end
```

konalegi: Thanks, yeah, looks better and we do lazy instantiation as before. Although in our code base I prefer to use `[]` hash access method, since it's called literally on each field and method_missing gives a performance penalty.

dgsuarez (approved):

barthez (approved):

PR #869, opened 2023-01-12
[feature] delayed_sidekiq strategy

skcc321:
# What does it do?
I added a new strategy - `delayed_sidekiq`.
It behaves the next way:
- the strategy has a `latency` option (for instance 10 seconds), `margin` & `reindex_wrapper`
- when you mutate active record objects very frequently the strategy will accumulate ids of mutated records during the latency window.
- that means that we are going to reindex all changes that happened during 10 seconds at once.
- such a latency gives us the opportunity to use "read replica" instead of "primary" DB (RoR multi DB connection)
- as a result, if you don't need strong consistency you can define an appropriate latency window and gain significant performance improvement (fewer calls to DB & replica usage)
- `margin` option is important to cover `replication_lag`

an example:
```ruby
class CitiesIndex < Chewy::Index
  index_scope City
  
  strategy_config delayed_sidekiq: {
    latency: 10,  
    margin: 2,
    reindex_wrapper: ->(&reindex) { ActiveRecord::Base.connected_to(role: :reading) { reindex.call }} # fetch data from read replica instead of the primary node
  }
  
  field :name
  field :address
  ...
end

class City < ActiveRecord::Model
  update_index("cities") { self if persistently_changed? }
  ...
end

------ Request 1 -------
Chewy.strategy(:delayed_sidekiq) do
  City.create(name: "Amsterdam") # created city with id=1
end
------ Request 2 -------
Chewy.strategy(:delayed_sidekiq) do
  City.create(name: "Paris") # created city with id=2
end
------ Request 3 -------
Chewy.strategy(:delayed_sidekiq) do
  City.create(name: "Rome") # created city with id=3
end
```

those three parallel requests that happened during the "latency" window are going to schedule a single worker to call import with all accumulated ids 1, 2, 3
```ruby
CityIndex.import([1, 2, 3]) # sql request will go to { role: :reading } due to definition in the index
```

## Strategy option in #import method.

```ruby
CityIndex.import([1, 2, 3], strategy: :delayed_sidekiq)
```

## Accumulation of :update_fields option between chunks.
```ruby
CityIndex.import([1, 2], update_fields: ["name"], strategy: :delayed_sidekiq) 
CityIndex.import([3], update_fields: ["address"], strategy: :delayed_sidekiq)
```
will cause
```ruby
CityIndex.import([1, 2, 3], update_fields: ["name", "address"])
```
that can be very useful if you have some complex crutches in the index definition and you don't want to load the database with obviously useless SQL queries.

## Configuration
[...]

PR #894, opened 2023-09-04
Add helper to cleanup delayed sidekiq timechunks

Drowze:
# For context:
https://github.com/toptal/chewy/pull/869 added the `delayed_sidekiq` strategy, which by design will only create a sidekiq worker to update an index every N seconds (10 seconds by default).
The implementation details for that involves redis sorted sets: it creates a sorted set for each index that uses the strategy and adds a new member to it everytime we call reindex (with the weight being calculated from the current timestamp, generating a unique number every N seconds).

# The problem:
Since we're writing directly to Redis, if we use this strategy in tests we're subject to state leaking and creating flaky tests. For example:
* test A runs some code that calls `MyClass.import!(1, strategy: :delayed_sidekiq)` and not flush the Sidekiq workers or the redis database
* in less than 10 seconds, test B runs some code that calls `MyClass.import!(1, strategy: :delayed_sidekiq)`, flushes Sidekiq workers and expects the index to be updated.  

On the above scenario, test B will **fail** as no worker will be enqueued since `import!` was called within 10 seconds in between the tests (also note that Sidekiq runs in memory by default when running tests, so workers enqueued by test A are not present in test B).

# The fix:
This PR adds a way to workaround the issue by cleaning up the saved timechunks. It can easily be added to e.g. rspec hooks via:
```ruby
RSpec.configure do |config|
  config.after { Chewy::Strategy::DelayedSidekiq.clear_timechunks! }
end
```

-----------------

* [ ] Feature branch is up-to-date with `master` (if not - rebase it).
* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### New Features
 
+* [#894](https://github.com/toptal/chewy/pull/894): Way of cleaning redis from artifacts left by `delayed_sidekiq` strategy which could potentially cause flaky tests. ([@Drowze](https://github.com/Drowze))
+
 ### Changes
 
 ### Bugs Fixed
--- a/README.md
+++ b/README.md
@@ -848,6 +848,12 @@ Explicit call of the reindex using `:delayed_sidekiq` strategy with `:update_fie
 CitiesIndex.import([1, 2, 3], update_fields: [:name], strategy: :delayed_sidekiq)
 ```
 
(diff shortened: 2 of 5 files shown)


konalegi:
Could you please rebase PR?

konalegi:
hey @Drowze , could you please rebase your PR?

Drowze:
Hey @konalegi, sorry for the delay! I didn't see the notification for the first message 😅 

Just rebased this PR! 👍

konalegi:
Hey, unfortunately rubocop is failing :( 
I suppose it's because we recently switched from ruby 2.x to 3.x syntax.

Drowze:
I see! Just fixed the rubocop issues and updated the PR 😄

konalegi (approved):

PR #911, opened 2023-12-12
Remove ruby 2.x

konalegi:
Clean-up everything ruby 2. x related, upgrade rubocop, and make it compliant with ruby 3.0 syntax.

-----------------

* [x] Added tests.


alejandroperea (approved):
Looks awesome! 👍

README.md
```diff
@@ -43,7 +43,7 @@ Or install it yourself as:
 
 ### Ruby
 
-Chewy is compatible with MRI 2.6-3.0¹.
+Chewy is compatible with MRI 3.0-3.2¹.
 
 > ¹ Ruby 3 is only supported with Rails 6.1
```
barthez, line 48: What does it mean?  Is it still relevant? Id Ruby 3.0/Rails 7 incompatible?

lib/chewy.rb
```diff
@@ -48,7 +48,7 @@ def try_require(path)
 require 'chewy/fields/base'
 require 'chewy/fields/root'
 require 'chewy/journal'
-require 'chewy/railtie' if defined?(::Rails::Railtie)
+require 'chewy/railtie' if defined?(Rails::Railtie)
```
barthez, line 51: What has changed we can skip the top module selector?

barthez (approved):

lib/chewy.rb
konalegi: This is what cop developer says 

> How Ruby searches constant is a bit complicated, and it can often be difficult to understand from the code whether the ‘::` is intended or not. Where `Module.nesting` is empty, there is no need to prepend `::`, so it would be nice to consistently avoid such meaningless `::` prefix to avoid confusion.

I've run all our specs on platform, and haven't seen any issues, so I suspect it's fine.

README.md
konalegi: ah, it means that ruby 3.0 starts support only with rails 6.1 :)

lib/chewy.rb
konalegi: Now I'm thinking maybe it makes sense to keep it, we cannot know why it has been added and considered gem is public, I'll bring it back

konalegi: Ok, I'm changing my opinion once again after careful inspection. In the same file for the previous version we `Rspec` and later `::Rspec`, so it seems there wasn't an intention to always use top-level consants.

PR #933, opened 2024-03-26
Relax allowed `elasticsearch` dependency version

mjankowski:
Saw the comment here - https://github.com/toptal/chewy/issues/847#issuecomment-1824503608 - which notes that newer 7.x versions of elasticsearch gem allow either Faraday 1.x or 2.x.

This change might be a good stopgap to include in a Chewy 7.x release while work on ES8 support (in Chewy 8) continues on that PR.

PR #947, opened 2024-05-01
Fix intermittent time-based failure in delayed sidekiq spec

mjankowski:
This example was previously failing when it was run at exactly an "even" 10 second increment time. For example, it would fail at "12:00:10", but pass at "12:00:09" and "12:00:11". This leads to intermittent failures on CI (presumably around ~10% of runs).

Updated to use a helper method in the spec which more closely mirrors the scheduler code.

I noticed this during the CI runs from the previous PR ... and was able to verify this pattern by setting a specific time target for the `Timecop.freeze` line, which confirmed the pattern of fail/pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Bugs Fixed
 
+* [#947](https://github.com/toptal/chewy/pull/947): Fix intermittent time-based failure in delayed sidekiq spec. ([@mjankowski][])
+
 ## 7.5.1 (2024-01-30)
 
 ### New Features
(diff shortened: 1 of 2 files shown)


konalegi:
It seems to be working, with no failure :wohoo: Could you please add the change log info? for instance like this https://github.com/toptal/chewy/pull/933/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR9

konalegi:
I'll trigger CI 4-5 times to make sure nothing fails, and once you add changelog I'll release it :)

mjankowski:
Added changelog commit here as well.

I see there may be a DIFFERENT intermittent failure in some of these CI runs (in rake helper spec), but it looks like none so far in the sidekiq one.

konalegi:
Yeah, it's different, related to sidekiq, seems to be not failing

konalegi (approved):

PR #973, opened 2024-11-24
Use suffixed index names by default to prevent alias creation conflicts

dmeremyanin:
## Synopsis

Recently, we encountered a problem during index reset operations where a race condition could cause an alias creation failure. The issue arises when:

1. An index (e.g., `users`) is being reset.
2. The reset task creates a new suffixed index (e.g., `users_<timestamp-suffix>`) and starts the import process.
3. Once the import is finished, the reset task deletes the old index ([here](https://github.com/toptal/chewy/blob/6a2176f7d92c4818f1a1f572aed339b39aef087e/lib/chewy/index/actions.rb#L166)).
4. The reset task attempts to create an alias (e.g., `users`) pointing to the new suffixed index ([here](https://github.com/toptal/chewy/blob/6a2176f7d92c4818f1a1f572aed339b39aef087e/lib/chewy/index/actions.rb#L167-L172)).
5. **The problem occurs**: There is a small time window between the old index deletion and the alias creation, during which another background job could recreate the unsuffixed `users` index. This results in a conflict when trying to create the alias, as both an index and an alias cannot share the same name.
6. **The result**: The alias creation fails, and no data is available in the index – the old index is deleted, and the new suffixed index is not aliased.

## Solution

To prevent this issue, I updated the index creation logic to always use suffixed index names and unsuffixed aliases by default, not just after an index reset. This ensures that the alias conflict issue caused by the recreation of the old unsuffixed index is avoided.

-----------------

* [x] Added tests.

### DIFF
--- a/lib/chewy/index.rb
+++ b/lib/chewy/index.rb
@@ -197,6 +197,16 @@ module Chewy
         public_methods - Chewy::Index.public_methods
       end
 
+      # Generates suffix for index names. By default, the method
+      # returns a timestamp-based suffix, using the current time in milliseconds (e.g., 1638947285000).
+      #
+      # It is used to differentiate indexes for zero-downtime deployments.
+      #
+      # @return [Object] an object that can be cast to a string (e.g., integer, timestamp, etc.)
+      def timestamp_suffix
+        (Time.now.to_f * 1000).round
+      end
+
       def settings_hash
         _settings.to_hash
       end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Changes
 
+* [#973](https://github.com/toptal/chewy/pull/973): **(Potentially Breaking)** Indexes now use suffixed physical names from first creation (e.g., `users_1700000000000`) with an unsuffixed alias (`users`), instead of switching to this layout only after an index reset. This reduces the risk of a race condition during index reset. Existing unsuffixed indexes are not auto-migrated and continue to work, but direct ES operations that expect unsuffixed concrete index names may need updates. To migrate existing unsuffixed indexes to the alias-based layout, perform a one-time `chewy:reset` in a controlled window (pause jobs/writes that can recreate the unsuffixed index name). ([@dmeremyanin][])
+
 ## 8.1.0 (2026-05-28)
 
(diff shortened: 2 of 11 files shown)


dmeremyanin:
@toptal-anvil ping reviewers

dmeremyanin:
@toptal-anvil ping reviewers

lib/chewy/index.rb
```diff
@@ -197,6 +197,16 @@ def scopes
         public_methods - Chewy::Index.public_methods
       end
 
+      # Generates an automatic suffix for index names. By default, the method
+      # returns a timestamp-based suffix, using the current time in milliseconds (e.g., 1638947285000).
+      #
+      # It is used to differentiate indexes for zero-downtime deployments.
+      #
+      # @return [Object] an object that can be cast to a string (e.g., integer, timestamp, etc.)
+      def auto_suffix
```
AlfonsoUceda: I am thinking to rename this method to `timestamp_suffix` to be more explicit with the purpose of the method

AlfonsoUceda:
Seems good @dmeremyanin thanks for the PR. I will need to test carefully and see if this can land in v8 or in the future v9

lib/chewy/index.rb
dmeremyanin: Done, renamed `auto_suffix` to `timestamp_suffix`

AlfonsoUceda:
@dmeremyanin after researching and reading the PR properly I think it is better to solve the non-atomic operation that happens in reset! action. That means in the update alias operation we remove also the index, that means this line https://github.com/toptal/chewy/pull/1021/changes#diff-d6f09601046ed2be417b8db0bd62e7c9c741aaf59fbae765de1de0f6b1819320L166 get replaced with https://github.com/toptal/chewy/pull/1021/changes#diff-d6f09601046ed2be417b8db0bd62e7c9c741aaf59fbae765de1de0f6b1819320R169-R172

I've also added the exists? change you did, worth it. What do you think?

https://github.com/toptal/chewy/pull/1021

AlfonsoUceda:
Closing this PR because we have decided to fix the bug with the solution in https://github.com/toptal/chewy/pull/1021

PR #996, opened 2026-01-29
Add context option to .import for passing custom data  

dimitrikochnev:
# Motivation

When importing objects, the data needed for indexing (e.g., embeddings, computed attributes) is often already available in memory at the call site. However, currently, crutch blocks are forced to re-fetch (even with direct_import) this data from the database, resulting in redundant queries and wasted resources.

There is currently no mechanism to pass this pre-computed data from the import call site down into crutch blocks or field value procs.

This design aligns with established patterns in the Ruby ecosystem, such as graphql-ruby and ActiveModelSerializers. In those libraries, a context object is similarly passed through the resolution or serialization stack to share request-level state (like current user, auth tokens, or pre-loaded data) without relying on global state.

# Solution

Add an optional `context:` keyword argument to `import`/`import!` that flows through the entire indexing pipeline. Context is an arbitrary hash, defaulting to `{}`.
```ruby
# Pass pre-computed data to avoid redundant DB queries
MyIndex.import!(objects, context: { embeddings: precomputed_embeddings })
```

## Context in crutch blocks (2nd argument)
```ruby
crutch :embeddings do |collection, context|
  # Use pre-computed data if available, otherwise fetch from DB
  context[:embeddings] || load_embeddings(collection)
end
```

## Context in field value procs (3rd argument)
```ruby
field :embedding, value: ->(object, crutches, context) {
  context[:override] || crutches.embeddings[object.id]
}
```

Both are fully backward-compatible — existing 1-arg crutch blocks and 1-2 arg field procs continue to work unchanged via arity-based dispatch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@
 
 ### New Features
 
+* [#996](https://github.com/toptal/chewy/pull/996): Add `context:` option to `import`/`import!` for passing custom data to crutch blocks and field value procs without redundant DB queries.
+
 ### Changes
 * [#977](https://github.com/toptal/chewy/pull/977): Fewer files on gem installation [@ericproulx](https://github.com/ericproulx).
 
--- a/lib/chewy/fields/root.rb
+++ b/lib/chewy/fields/root.rb
@@ -61,8 +61,8 @@ module Chewy
       # @param fields [Array<Symbol>] a list of fields to compose, every field will be composed if empty
       # @return [Hash] JSON-ready hash with stringified keys
       #
-      def compose(object, crutches = nil, fields: [])
-        result = evaluate([object, crutches])
+      def compose(object, crutches = nil, fields: [], context: {})
+        result = evaluate([object, crutches, context])
 
         if children.present?
           child_fields = if fields.present?
@@ -72,7 +72,7 @@ module Chewy
           end
 
           child_fields.each_with_object({}) do |field, memo|
-            memo.merge!(field.compose(result, crutches) || {})
+            memo.merge!(field.compose(result, crutches, context) || {})
           end.as_json
         elsif fields.present?
           result.as_json(only: fields, root: false)
--- a/lib/chewy/index/crutch.rb
+++ b/lib/chewy/index/crutch.rb
@@ -9,9 +9,12 @@ module Chewy
       end
 
       class Crutches
-        def initialize(index, collection)
+        attr_reader :context
+
+        def initialize(index, collection, context = {})
           @index = index
           @collection = collection
+          @context = context
           @crutches_instances = {}
         end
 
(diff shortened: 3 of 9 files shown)


dimitrikochnev:
Let me find there is it :-)

dimitrikochnev:
Done. 

I will give you a practical example storing video embeddings:

```ruby
class Video::EncodeSemanticProcessor < ProcessorMan::Processor
  include MediaProcessing

  option :model, :string, default: "auto"

  def process
    upstream_media, upstream_video = require_upstreams(0 => :media, 1 => :video)

    result = with_progress(0.1, 0.9) do |progress|
      Ferment::Engine.stream("/v1/video/encode-semantic", video_in: upstream_video.cook.blob.url, **options.to_params) do |event|
        progress.call(event[:progress]) if event[:type] == :progress
      end
    end

    MultiTenancy.with(network.user) do
      visual_slices, base64_embeddings = build_visual_slices(result)

      media.update!(visual_slices: visual_slices)

      Media::Embedding.bulk_upsert(media, base64_embeddings)  # <--- Store raw embeddings in the db

      Chewy.strategy(:atomic) do
        context = { embeddings: Media::Embedding.to_floats(base64_embeddings) }

        MediaIndex.delete_for(media, type: 'Media::Data::VisualSlice')
        MediaIndex.import!(Array.wrap(media.visual_slices), direct_import: true, context: context) # <--- Don't perform a huge round trip to the db, classic reindex still works since the embeddings will be fetched from the origin db
      end
    end

    succeed(extract_data(result, strip_nested: true))
  end

  private

    def build_visual_slices(result)
      embeddings = {}

      slices = result.fetch(:slices).map do |slice_data|
        slice = Media::Data::VisualSlice.new(
          id: media_cid(:visual_slice, slice_data.fetch(:position)),
          blob_id: media.id,
          blob_type: media.class.name,
          user_id: media.user_id,
          position: slice_data.fetch(:position),
          start_time: slice_data.fetch(:start_time),
          duration: slice_data.fetch(:duration),
          first_frame: slice_data.fetch(:first_frame),
          frame_count: slice_data.fetch(:frame_count),
          semantic_embedding_model: result.fetch(:model), 
          semantic_embedding_dim: result.fetch(:embedding_dim),
        )

        embeddings[slice.embedding_key] = slice_data.fetch(:embedding) # <--- HUGE

        slice
      end

      [slices, embeddings]
    end
end
```

AlfonsoUceda:
@dimitrikochnev thank you for your contribution. Could you rebase master onto your branch? Would like to run CI.

dimitrikochnev:
@AlfonsoUceda hi! done!

AlfonsoUceda (approved):
Going to fix rubocop problem and there is also a fix needed for parallel version. We cannot delete(:context) from the options

PR #1009, opened 2026-03-17
Add Elasticsearch 9.x support

mattmenefee:
> **Note:** This PR depends on #1008 (CI and test infrastructure improvements) and #1010 (migration guide and doc fixes) which should be merged first.
>
> **Related:** #1028 removes the obsolete `_type` field (mock helpers, `EVERFIELDS`, `Index::Wrapper` accessors).

## Summary

- Widen `elasticsearch` gem dependency from `< 9.0` to `< 10.0` to support ES 9.x clusters
- Restructure scroll API calls to nest `scroll_id` under `body:`, matching the elasticsearch-ruby 9.x client interface
- Use `.body` accessor on `cat.indices` response for ES 9.x gem compatibility
- Pin `elasticsearch` gem to `~> 8.14` in CI for ES 8.x matrix entries (the 9.x gem sends a `compatible-with=9` header that ES 8.x rejects)
- Add ES version matrix to CI: 9.4.1 default with three 8.19.3 spot-check entries (Ruby 3.3/3.4/4.0)
- Add a Chewy 8→9 migration guide section with code examples, gem pinning guidance, and breaking-change notes
- Add a Chewy 9.0.x (ES 8.x + 9.x) row to the README compatibility table

### Breaking changes

- The `search_query.chewy` notification payload for scroll requests now nests `scroll_id` under `body:` (i.e. `{scroll: '1m', body: {scroll_id: '...'}}` instead of `{scroll: '1m', scroll_id: '...'}`). Update any application code that subscribes to scroll notification payloads.

## Test plan

- [x] Verify all existing specs pass against ES 8.x
- [x] Verify all existing specs pass against ES 9.x
- [x] Verify scroll pagination works end-to-end on both ES versions
- [x] Verify `search_query.chewy` notification payload matches new format

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -27,7 +27,21 @@ jobs:
       matrix:
         ruby: [ '3.2', '3.3', '3.4', '4.0' ]
         gemfile: [rails.7.2.activerecord, rails.8.0.activerecord, rails.8.1.activerecord]
-    name: ${{ matrix.ruby }}-${{ matrix.gemfile }}
+        elasticsearch: ['9.4.3']
+        include:
+          - ruby: '3.3'
+            gemfile: rails.7.2.activerecord
+            elasticsearch: '8.19.3'
+            es_gem_constraint: '~> 8.14'
+          - ruby: '3.4'
+            gemfile: rails.8.0.activerecord
+            elasticsearch: '8.19.3'
+            es_gem_constraint: '~> 8.14'
+          - ruby: '4.0'
+            gemfile: rails.8.1.activerecord
+            elasticsearch: '8.19.3'
+            es_gem_constraint: '~> 8.14'
+    name: ${{ matrix.ruby }}-${{ matrix.gemfile }}-es${{ matrix.elasticsearch }}
 
     env:
       BUNDLE_GEMFILE: gemfiles/${{ matrix.gemfile }}.gemfile
@@ -45,7 +59,7 @@ jobs:
           --health-timeout 5s
           --health-retries 5
       elasticsearch:
-        image: docker.elastic.co/elasticsearch/elasticsearch:8.19.13
+        image: docker.elastic.co/elasticsearch/elasticsearch:${{ matrix.elasticsearch }}
         ports:
           - '9250:9200'
         env:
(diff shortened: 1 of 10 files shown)


bbatsov (commented):
Thanks for tackling ES 9 support! We'll review this properly after #1008 and #1010 have been updated and merged, since this builds on top of both.

One thing I'd suggest splitting out now: the Sidekiq testing API changes (`require 'sidekiq/testing'` -> `Sidekiq.testing!(:fake)`, `Sidekiq::Testing.inline!` -> `Sidekiq.testing!(:inline)`) are unrelated to ES 9 and would be easier to review and merge as a separate PR.

.github/workflows/ruby.yml
```diff
@@ -27,7 +27,21 @@ jobs:
       matrix:
         ruby: [ '3.2', '3.3', '3.4', '4.0' ]
         gemfile: [rails.7.2.activerecord, rails.8.0.activerecord, rails.8.1.activerecord]
-    name: ${{ matrix.ruby }}-${{ matrix.gemfile }}
+        elasticsearch: ['9.3.2']
```
mattmenefee, line 30: Would it be better / cleaner to use something like [appraisal](https://github.com/thoughtbot/appraisal) to test against both Elasticsearch v8 and v9?

.github/workflows/ruby.yml
```diff
@@ -27,7 +27,21 @@ jobs:
       matrix:
         ruby: [ '3.2', '3.3', '3.4', '4.0' ]
         gemfile: [rails.7.2.activerecord, rails.8.0.activerecord, rails.8.1.activerecord]
-    name: ${{ matrix.ruby }}-${{ matrix.gemfile }}
```
mattmenefee, line 30: @AlfonsoUceda Would it be better / cleaner to use something like [appraisal](https://github.com/thoughtbot/appraisal) to test against both Elasticsearch v8 and v9?

.github/workflows/ruby.yml
```diff
@@ -27,7 +27,21 @@ jobs:
...
+        elasticsearch: ['9.4.1']
+        include:
+          - ruby: '3.2'
```
mattmenefee: This should be adjusted to `3.3` if #1029 is merged

mattmenefee:
@AlfonsoUceda I've been rebasing this onto `master` periodically as new commits have landed. Do you want me to keep doing that until this PR is merged?

I've mainly been doing this because I'm running this branch in Production, and picking up the new changes/improvements on `master` in addition to the upgrade to ES v9 has been convenient.

PR #1029, opened 2026-06-01
Drop support for Ruby 3.2 and remove dead version guards

mattmenefee:
## Summary

- **(Breaking)** Drops support for Ruby 3.2, which reached [end-of-life on 2026-04-01](https://www.ruby-lang.org/en/downloads/branches/) and no longer receives security patches.
- Raises the floor to **Ruby 3.3** (currently in security maintenance, EOL expected 2027-03-31):
  - `chewy.gemspec`: `required_ruby_version` `>= 3.2` → `>= 3.3`
  - `.github/workflows/ruby.yml`: drop `'3.2'` from the test matrix (3.3 was already covered)
  - `.rubocop.yml`: `TargetRubyVersion` `3.2` → `3.3`
  - README compatibility statement and CONTRIBUTING prerequisites
  - Troubleshooting guide: drop the `parser` / `prism` version conditional from the Witchcraft optional-dependency list, leaving `method_source` and `unparser` (Prism is bundled with Ruby 3.3+ and `parser` is pulled in transitively by `unparser`, so neither needs adding to a Gemfile)

### Dead version-guard cleanup (exposed by the floor bump)

- **`Chewy::Index::Witchcraft`** branched on `RUBY_VERSION >= '3.3'` to choose between **Prism** (a default gem since Ruby 3.3) and the external `parser` gem. The `parser` fallback in the `require` block, `check_requirements!`, and `ast_from_proc` is now unreachable, so witchcraft always uses Prism — also removing a fragile *string* version comparison. The Prism requirements check itself is dropped: Prism ships with Ruby 3.3+, so it's always present and never needs adding to a Gemfile; the real `parser` dependency is already covered transitively by the `unparser` check.
- **`Chewy::LogSubscriber`** branched on `ActiveSupport.version >= 7.1`, unconditionally true since the AS `>= 7.2` floor; the pre-7.1 positional `color` call is removed.

Version-anchored docs describing *past* releases' requirements (e.g. the Chewy 7 → 8 migration guide) are intentionally left unchanged.

## Test plan

- [x] `bundle exec rubocop` — no offenses with `TargetRubyVersion: 3.3`
- [x] `gem build chewy.gemspec` succeeds with `required_ruby_version >= 3.3`
- [ ] CI green on the Ruby 3.3 / 3.4 / 4.0 matrix

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [ '3.2', '3.3', '3.4', '4.0' ]
+        ruby: [ '3.3', '3.4', '4.0' ]
         gemfile: [rails.7.2.activerecord, rails.8.0.activerecord, rails.8.1.activerecord]
     name: ${{ matrix.ruby }}-${{ matrix.gemfile }}
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,7 +5,7 @@ plugins:
 
 AllCops:
   NewCops: enable
-  TargetRubyVersion: 3.2
+  TargetRubyVersion: 3.3
   SuggestExtensions: false
 
 Layout/AccessModifierIndentation:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Changes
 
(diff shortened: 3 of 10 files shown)


AlfonsoUceda:
Thanks @mattmenefee I'm planning this for v9

Issue #644, opened 2018-05-01
Witchcraft memory allocation


Gargron:
I've run my Rails app through memory_profiler. There is one Chewy index definition that does not use the "witchcraft" technology. Nevertheless, Chewy comes up at the top of memory allocating gems with 13.6797 MiB, with the witchcraft dependency graph being the most of it:

```
    $HOME_DIR/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/chewy-5.0.0/lib/chewy.rb: 13.6797 MiB
      $HOME_DIR/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/chewy-5.0.0/lib/chewy/type.rb: 7.3242 MiB
        $HOME_DIR/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/chewy-5.0.0/lib/chewy/type/witchcraft.rb: 7.3242 MiB
          $HOME_DIR/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/current.rb: 7.3242 MiB
            $HOME_DIR/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/ruby25.rb: 7.3242 MiB
              $HOME_DIR/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser.rb: 5.2656 MiB
                $HOME_DIR/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/parser-2.5.1.0/lib/parser/lexer.rb: 5.2656 MiB
      $HOME_DIR/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/chewy-5.0.0/lib/chewy/index.rb: 3.3047 MiB
        $HOME_DIR/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/chewy-5.0.0/lib/chewy/search.rb: 3.3047 MiB
          $HOME_DIR/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/gems/chewy-5.0.0/lib/chewy/search/parameters.rb: 2.7891 MiB
[...]

PR #1032, opened 2026-06-03
Compiled compose path replaces witchcraft! as the default

AlfonsoUceda:
`witchcraft!` made imports fast by reading each field's value proc source, parsing it, and emitting one fused lambda per index — but it was opt-in, fragile (depended on `method_source` reflowing proc source correctly), and dragged `parser`/`prism`/`unparser` into every Rails boot whether or not an index used it (~22 MiB of allocations, ~0.9 MiB retained per process, see #644). This replaces it with a compiler that builds the same single-method compose path straight from the field tree, so every index gets witchcraft-equivalent throughput by default with zero dependencies and nothing to opt into.

On the first import Chewy generates one `__chewy_compose__` method per index and reuses it for every object, replacing the iterative `Fields::Root#compose` loop. `update_fields:` partial imports stay on the fast path via a separate memoized method per unique fields set. A handful of cases the compiler can't express (`ignore_blank`, `geo_point`, custom root value procs, non-identifier field names) fall back to the legacy iterative path transparently, producing identical output.

`witchcraft!` is deprecated, not removed: it still works, but now prints a deprecation warning and loads its parser dependencies lazily, so apps that never call it pay no boot cost. Removing the call switches an index to the compiled path with no other change.

## Benchmarks

Measured locally on Ruby 4.0.4 with the scripts under `benchmark/` (CPU: 10k objects, median of 5 runs; memory via `memory_profiler`). The correctness harness confirms plain, witchcraft, and compiled produce byte-identical documents across all field tiers.

**Composition throughput** — speedup vs the iterative path:

| fields | iterative | witchcraft | compiled | compiled vs witch |
|-------:|----------:|-----------:|---------:|------------------:|
| 5      | 1.00×     | 4.98×      | 5.16×    | 1.04×             |
| 20     | 1.00×     | 4.26×      | 4.31×    | 1.01×             |
| 50     | 1.00×     | 4.30×      | 4.31×    | 1.00×             |

Compiled matches witchcraft within ~1% and is 4–5× faster than the iterative loop.

**Memory** — compiled vs witchcraft:

| measurement                    | compiled        | witchcraft       |
|--------------------------------|-----------------|------------------|
| boot + first compose, allocated | 0.49 MB         | 22.2 MB          |
| boot + first compose, retained  | ~0              | 0.93 MB          |
| steady-state per compose        | 2344 B / 11 obj | 3144 B / 16 obj  |

Dropping the parser stack from boot saves ~21.7 MB of allocations and ~0.9 MB retained per process for the common case. Compiled is also lighter per compose at steady state.

## Testing

- `bundle exec rspec spec/chewy/index/compiled_spec.rb` — covers compiled output parity, `fields:` restriction, proc-arity dispatch, and each fallback case.
- `bundle exec ruby benchmark/witchcraft_bench.rb` and `ruby benchmark/witchcraft_memory.rb` — reproduce the numbers above (the memory script needs the `memory_profiler` gem).
- Run an existing index import with and without `witchcraft!`; documents are identical and the deprecation warning fires only when `witchcraft!` is called.

-----------------

* [x] Added tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ tmp
 _site
 .sass-cache
 file::memory:?cache=shared
+benchmark
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,13 @@
 
 ### New Features
 
+* New compiled compose path is now the default for every index. Chewy generates one `__chewy_compose__` method per index from the field tree on first import and reuses it for every object, replacing the iterative `Fields::Root#compose` loop. Typical document composition is 3-4× faster than the iterative path and matches `witchcraft!` performance without any extra dependencies. `update_fields:` partial imports are also covered via per-fields-set memoized methods.
+
 ### Bug Fixes
 
 ### Changes
+
+* **Deprecation:** `Chewy::Index.witchcraft!` is deprecated and will be removed in a future major release. It now prints a deprecation warning and is no longer required for fast composition — the compiled compose path described above is the default and delivers equivalent throughput without `method_source` / `parser` / `prism` / `unparser`. The `parser` / `prism` / `unparser` requires are now lazy and only loaded when an index actually calls `witchcraft!`, eliminating ~24 MiB of boot-time allocations and ~1 MiB of retained memory for apps that don't use it ([#644](https://github.com/toptal/chewy/issues/644)).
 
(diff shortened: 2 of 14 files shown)


bbatsov (approved):

bbatsov:
Well done!

PR #1043, opened 2026-06-18
[Fix #1042] Symbol#to_proc field values on the compiled compose path are corrected

### LINKED ISSUE #1042: `value: proc(&:method)` fields broken in 8.4.0 (worked in 8.0.1)
After upgrading to 8.4.0, indexing fails for fields defined with a `Symbol#to_proc`. This worked on 8.0.1.
--------

## Expected behavior

`proc(&:method)` fields keep working as in 8.0.1.

## Actual behavior & steps to reproduce the problem

```
  class MyIndex < Chewy::Index
    index_name :my_index
    field :published, value: proc(&:published?)
  end
  # importing a record raises:
  # ArgumentError: wrong number of arguments (given 2, expected 0)
```

## Version Information

* Chewy version 8.4
* Elasticsearch version 8.19.3
* Ruby version 4.0.5
* Rails version 8.1

AlfonsoUceda:
Fixes #1042.

Indexing a field defined with the `Symbol#to_proc` shorthand (`value: proc(&:published?)`) raised `ArgumentError: wrong number of arguments (given 2, expected 0)` on 8.4.0, having worked on 8.0.1. The regression came from the compiled compose path introduced in 8.4.0: for any non-zero-arity proc it forwarded `(object, crutches, context)` as positional arguments. A `Symbol#to_proc` declares a splat, so it then passed `crutches` and `context` straight to the method (`object.published?(crutches, context)`). The legacy/fallback `Fields::Base#value_by_proc` never did this, which is why the plain path stayed correct.

The compiler now special-cases splat-declaring procs — `Symbol#to_proc` and `->(*args)` — and calls them with the object alone, matching the plain compose path exactly. Splat-free procs keep the existing `(object, crutches, context)` truncation, so optional-argument lambdas like `->(o, c = nil)` are unaffected.

## Testing

- `bundle exec rspec spec/chewy/index/compiled_spec.rb` — green, including the new `value: Symbol#to_proc` case. The test uses a real object rather than an RSpec double, since a double accepts any argument count and would mask the bug.

-----------------

* [x] Added tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Bug Fixes
 
+* Fixed `field ..., value: proc(&:method)` (and other splat-declaring procs) raising `ArgumentError: wrong number of arguments` during import. The compiled compose path introduced in 8.4.0 forwarded `crutches`/`context` as extra positional arguments to splat procs; `Symbol#to_proc` then passed them on to the method. Splat procs are now called with the object alone, matching the plain compose path. ([@AlfonsoUceda][])
+
 ### Changes
 
 ## 8.4.0 (2026-06-17)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,15 @@
 
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


andreyruby:
Yeah, all green on this branch 👍

bbatsov (approved):

